Issue #11, opened 2018-02-26
Introspection


crisu83:
We need to add basic support for introspection queries.

crisu83:
Basic support added in #15.

PR #15, opened 2018-02-26
Initial support for introspection queries

crisu83:
Gave introspection a shot, but it seems that the execution part cannot handle the query yet.

What would be the best way to solve this @hungneox?


crisu83:
Fixes #11.

PR #19, opened 2018-02-27
Initial support for building schema from SDL

crisu83:
We need support for building an executable schema from Schema Definition Language (SDL).


src/Language/buildSchema.php
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * @param string $source
+ * @param array  $options
+ * @return SchemaInterface
+ * @throws \Digia\GraphQL\Error\GraphQLError
+ * @throws \Exception
+ * @throws \TypeError
+ */
+function buildSchema(string $source, array $options = []): SchemaInterface
```
hungneox: Should it be `BuildSchema` for consistency?

src/Type/Definition/Behavior/ArgumentsTrait.php
```diff
@@ -12,62 +12,38 @@
     /**
      * @var Argument[]
      */
-    private $args = [];
+    private $_arguments = [];
```
hungneox: Personally I don't think underscore is really necessary

src/Language/buildSchema.php
hungneox: I mean since we have `GraphQLSchema`, `GraphQLXXX` function. In Laravel functions should be in snake_case :-?

crisu83: Usually, functions are named using camelCase in PHP. GraphQLXXX is an exception to this, which I made because it makes it look more consistent with the other implementations.

src/Type/Definition/Behavior/ArgumentsTrait.php
crisu83: The underscore just signifies that it cannot be set directly from outside the class. This is more of a convention I came up with to maintain my sanity.

PR #20, opened 2018-02-28
Flatten directory structure

crisu83:


crisu83:
I didn't touch the `Execution` so this should be fine to merge.

PR #21, opened 2018-03-02
Dependecy injection

crisu83:
Clean up the tight coupling between the classes by introducing a GraphQL runtime that uses dependency injection to wire everything together.


composer.json
```diff
@@ -34,5 +34,6 @@
         "psr-4": {
             "Digia\\GraphQL\\Test\\": "./tests"
         }
-    }
+    },
+    "minimum-stability": "dev"
```
Jalle19: Why?

Issue #12, opened 2018-02-26
Print AST tree


crisu83:
We need to implement support for converting AST trees into GraphQL strings in order to implement #11. We have already implemented all of the nodes so this should be fairly straightforward. 

The reference implementation can be found here:
https://github.com/graphql/graphql-js/blob/master/src/language/printer.js

crisu83:
This is not a good first issue because we need to solve the visitor part first.

crisu83:
Partially done in #51.

crisu83:
This has been done a long time ago.

Issue #3, opened 2018-02-15
Validation rules


crisu83:
We need to add support for validation using the Visitor pattern.

The following rules have been implemented:
- [x] ExecutableDefinitions
- [x] FieldOnCorrectType
- [x] FragmentsOnCompositeTypes
- [x] KnownArgumentNames
- [x] KnownDirectives
- [x] KnownFragmentNames
- [x] KnownTypeNames
- [x] LoneAnonymousOperation
- [x] NoFragmentCycles
- [x] NoUndefinedVariables
- [x] NoUnusedFragments
- [x] NoUnusedVariables
- [x] OverlappingFieldsCanBeMerged
- [x] PossibleFragmentSpreads
- [x] ProvidedNonNullArguments
- [x] ScalarLeafs
- [x] SingleFieldSubscriptions
- [x] UniqueArgumentNames
- [x] UniqueDirectivesPerLocation
- [x] UniqueFragmentNames
- [x] UniqueInputFieldNames
- [x] UniqueOperationNames
- [x] UniqueVariableNames
- [x] ValuesOfCorrectType
- [x] VariablesAreInputTypes
- [x] VariablesDefaultValueAllowed
- [x] VariablesInAllowedPosition

crisu83:
Added in #45.

Jalle19:
Should we close this?

PR #23, opened 2018-03-02
WIP: Add initial support for visitors

crisu83:
We need to add support for visiting an AST tree in order to implement #3 and #12.

### DIFF
--- a/src/Config/ConfigTrait.php
+++ b/src/Config/ConfigTrait.php
@@ -23,6 +23,35 @@ trait ConfigTrait
     }
 
     /**
+     * @param string $key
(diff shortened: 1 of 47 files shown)


crisu83:
@hungneox it seems that the test you added fails. I probably found the issue, it was this bug in the `FragmentDefinitionBuilder`:
https://github.com/digiaonline/graphql-php/pull/23/files?diff=unified#diff-eee9b4abbf7a321963db300bf5f49cd5R17

crisu83:
It's getting there, but we're still missing support for visiting in parallel and visiting with type info.

PR #25, opened 2018-03-05
Execution fragments deeper

hungneox:
I have to improve the implementation a lot more but basically now the execution can execute multiple fragments.
It's done in a recursive way which I think not so good. I will try to do it in non-recursive way.

### DIFF
--- a/src/Execution/ExecutionStrategy.php
+++ b/src/Execution/ExecutionStrategy.php
@@ -71,7 +71,7 @@ abstract class ExecutionStrategy
             /** @var FieldNode $selection */
             switch ($selection->getKind()) {
                 case NodeKindEnum::FIELD:
-                    $name = $selection->getNameValue();
+                    $name = $this->getFieldNameKey($selection);
                     if (!isset($fields[$name])) {
                         $fields[$name] = new \ArrayObject();
                     }
@@ -108,6 +108,15 @@ abstract class ExecutionStrategy
     }
 
     /**
+     * @param FieldNode $node
+     * @return string
+     */
+    private function getFieldNameKey(FieldNode $node)
+    {
+        return $node->getAlias() ? $node->getAlias()->getValue() : $node->getNameValue();
+    }
+
+    /**
      * Implements the "Evaluating selection sets" section of the spec
      * for "read" mode.
      * @param ObjectType $parentType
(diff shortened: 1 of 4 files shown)


src/Execution/ExecutionStrategy.php
```diff
@@ -107,6 +116,17 @@ protected function collectFields(
         return $fields;
     }
 
+    /**
+     * @param FieldNode $node
+     * @return string
+     */
+    private function getFieldNameKey(FieldNode $node)
+    {
+        return $node->getAlias()
```
crisu83: We should a `hasAlias` method to the associated trait and use that instead.

src/Execution/ExecutionStrategy.php
```diff
@@ -130,12 +150,9 @@ protected function executeFields(
         foreach ($fields as $fieldName => $fieldNodes) {
             $fieldPath   = $path;
             $fieldPath[] = $fieldName;
-            if (!$this->isDefinedField($parentType, $fieldName)) {
-                continue;
-            }
 
             $result = $this->resolveField($parentType,
-                $source,
+                [],
```
crisu83: We don't support arguments yet?

src/Execution/ExecutionStrategy.php
```diff
@@ -173,38 +179,47 @@ protected function resolveField(
...
+                );
+
+                $data = $this->executeFields(
+                    $parentType,
+                    $source,
+                    $path,
+                    $fields
+                );
+
+                $subResult = array_merge_recursive(
```
crisu83: This can be a costy operation. We should add a TODO to benchmark it.

src/Execution/ExecutionStrategy.php
```diff
@@ -173,38 +179,47 @@ protected function resolveField(
...
-        }
 
-        $result = $field->resolve(...$args);
+            $subResult = $field->resolve(...$args);
+
+            if (!empty($fieldNode->getSelectionSet())) {
+                $fields = $this->collectFields(
+                    $parentType,
+                    $fieldNode->getSelectionSet(),
+                    new \ArrayObject(),
```
crisu83: Why not use `[]` instead?

src/Execution/ExecutionStrategy.php
hungneox: We do support args 
```
$inputValues = $fieldNode->getArguments() ?? [];
            $args        = [];

            foreach ($inputValues as $value) {
                if ($value instanceof ArgumentNode) {
                    $args[] = $value->getValue()->getValue();
                } elseif ($value instanceof InputValueDefinitionNode) {
                    $args[] = $value->getDefaultValue()->getValue();
                }
            }
```

src/Execution/ExecutionStrategy.php
hungneox: I'm going to refactor this very soon :D

PR #27, opened 2018-03-05
Fix code inspection issues

crisu83:


src/Language/AST/Schema/SchemaBuilder.php
```diff
@@ -132,8 +132,9 @@ function getOperationTypes(SchemaDefinitionNode $schemaDefinition, array $nodeMa
 
     foreach ($schemaDefinition->getOperationTypes() as $operationTypeDefinition) {
         $operationType = $operationTypeDefinition->getType();
-        $typeName      = $operationType->getNameValue();
-        $operation     = $operationTypeDefinition->getOperation();
+        /** @noinspection PhpUndefinedMethodInspection */
```
Jalle19: A `@var` annotation to help the IDE would probably be better than suppressing the warnings

PR #33, opened 2018-03-07
Start using Phpstan for static analysis

Jalle19:
It finds a lot of errors so I configured Travis CI to ignore failures from this step for now

### DIFF
--- a/src/Util/coercers.php
+++ b/src/Util/coercers.php
@@ -1,0 +1,93 @@
+<?php
+
+namespace Digia\GraphQL\Util;
+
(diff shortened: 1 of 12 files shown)


src/Util/coercers.php
```diff
@@ -0,0 +1,93 @@
+<?php
```
crisu83: Maybe we should move this logic into `Coercer` classes.

Jalle19: I figured since they're all static functions they could live in a function file

PR #37, opened 2018-03-07
Bump phpstan analysis level to 2

Jalle19:
@crisu83 @hungneox I will need your help nailing these last issues. Most of them come from the fact that traits are used in type-hints instead of interfaces.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,7 @@ before_script:
   - composer self-update
   - composer install
 
-script:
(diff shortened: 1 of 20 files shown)


crisu83:
Let's not do this before we have the first version ready. After that I'm all for bumping the level to 2.

Jalle19:
@crisu83 the things it complains about on level 2 are things that are better to fix before a "first version" since it will probably require some interface changes etc.

Jalle19:
Closing for now

PR #40, opened 2018-03-08
Pass environment to resolve function

hungneox:
screen shot 2018-03-08 at 4 27 36

@crisu83 I'm not sure if this is an expected behaviour or not but I think there should be `name` argument?


crisu83:
@Jalle19 Phpstan is failing all of our builds, can we drop the level back to 1?

crisu83:
@hungneox Arguments are lazy so the arguments should be under `$config` until it's actually invoked throught `getArguments`. Was this ok is there still an issue with this?

crisu83:
Rebased on *master*.

Jalle19:
Maybe we could add StyleCI to this project?

PR #43, opened 2018-03-09
Execution context builder

hungneox:
- Introduce Execution context builder
- Port some tests from `graphql-js` library
- Fix minor styling issues


src/Execution/Execution.php
```diff
@@ -42,12 +43,15 @@ public static function execute(
...
         $rootValue = null,
         $contextValue = null,
-        $variableValues = null,
+        $variableValues = [],
         $operationName = null,
         callable $fieldResolver = null
     ) {
         try {
-            $context = self::buildExecutionContext(
+            //@TODO Get context builder from container?
```
crisu83: Yes.

src/Execution/ExecutionContextBuilder.php
```diff
@@ -0,0 +1,79 @@
...
+        callable $fieldResolver = null
+    ): ExecutionContext {
+        //@TODO validate $rawVariableValues?
+
+        $errors    = [];
+        $fragments = [];
+        $operation = null;
+
+        foreach ($documentNode->getDefinitions() as $definition) {
+            switch ($definition->getKind()) {
```
crisu83: I would use `instanceof` instead.

src/Execution/ExecutionStrategy.php
```diff
@@ -172,27 +172,31 @@ protected function executeFields(
...
-     * @return \Digia\GraphQL\Type\Definition\Field|null
+     * @return Field|null
      * @throws \Exception
      * @throws \TypeError
      */
     public function getFieldDefinition(Schema $schema, ObjectType $parentType, string $fieldName)
     {
-        if ($fieldName === SchemaMetaFieldDefinition()->getName() && $schema->getQuery() === $parentType) {
-            return SchemaMetaFieldDefinition();
+        $schemaMetaFieldDifinition   = SchemaMetaFieldDefinition();
```
crisu83: I would move these just before the associated `if` -statement because we're using early return.

PR #49, opened 2018-03-09
Execution service provider

hungneox:
+ Add ExecutionInterface and ExecutionProvider
+ Fix all of execution tests


src/Provider/ExecutionProvider.php
```diff
@@ -0,0 +1,31 @@
...
+    protected $provides = [
+        ExecutionInterface::class
+    ];
+
+    /**
+     * @inheritdoc
+     */
+    public function register()
+    {
+        $this->container->add(ExecutionInterface::class, function () {
```
crisu83: Please change the second argument to `Execution::class`. Also, should this be a shared instance? If so, give `true` as the third argument.

src/facades.php
```diff
@@ -101,7 +102,7 @@ function graphql(
     callable $fieldResolver = null
 ): ExecutionResult {
     /** @noinspection PhpParamsInspection */
-    return Execution::execute(
+    return executor()->execute(
```
crisu83: I would prefer `GraphQL::get(ExecutionInterface::class)->execute(` here.

src/facades.php
```diff
@@ -111,3 +112,11 @@ function graphql(
         $fieldResolver
     );
 }
+
+/**
+ * @return ExecutionInterface
+ */
+function executor(): ExecutionInterface
```
crisu83: I think this method is unnecessary.

tests/Functional/Execution/ExecutionTest.php
```diff
@@ -60,12 +62,12 @@ public function testNoSchemaIsProvided()
     {
         $this->expectException(\TypeError::class);
 
-        /** @var ExecutionResult $executionResult */
-        Execution::execute(null, parse('{field}'));
+        graphql(null, parse('{field}'));
```
crisu83: Looks good! 👍

tests/Functional/Execution/ExecutionTest.php
```diff
@@ -84,17 +86,20 @@ public function testAcceptAnObjectWithNamedPropertiesAsArguments()
...
         ]);
 
-        $rootValue    = 'rootValue';
-        $documentNode = parse('query Example { a }');
+        $rootValue = 'rootValue';
+        $document  = parse('query Example { a }');
 
         /** @var ExecutionResult $executionResult */
-        $result = Execution::execute($schema, $documentNode, $rootValue);
+        $result = executor()->execute($schema, $document, $rootValue);
```
crisu83: I would prefer `$this->execution = GraphQL::get(ExecutionInterface::class);` inside the `setUp` method.

src/facades.php
hungneox: My test simply doesn't work with function `graphql(SchemaInterface $schema, string $source,...`. It looks unnecessary, but it's not. Let's I think a bit.

tests/Functional/Execution/ExecutionTest.php
hungneox: It's suppose not to work, the second parameter should be string . which is weird....

tests/Functional/Execution/ExecutionTest.php
hungneox: I agree, it's easier to change

src/facades.php
hungneox: Technically I can call only `graphql(...)` but for some reason all the tests failed.

src/helpers.php
```diff
@@ -96,18 +97,56 @@ function graphql(
...
-    return Execution::execute(
-        $schema,
-        parse($source),
-        $rootValue,
-        $contextValue,
-        $variableValues,
-        $operationName,
-        $fieldResolver
-    );
+    return GraphQL::getInstance()
```
crisu83: This is supposed to be `GraphQL::get(...)`.

tests/Functional/Execution/ExecutionTest.php
crisu83: Shouldn't this be `graphql(null, '{field}')`?

tests/Functional/Execution/ExecutionTest.php
```diff
@@ -84,17 +85,20 @@ public function testAcceptAnObjectWithNamedPropertiesAsArguments()
                 ])
         ]);
 
-        $rootValue    = 'rootValue';
-        $documentNode = parse('query Example { a }');
+        $rootValue = 'rootValue';
+        $document  = 'query Example { a }';
```
crisu83: Ain't this `$source`?

src/helpers.php
hungneox: Lol, I copied your code above...

tests/Functional/Execution/ExecutionTest.php
hungneox: It's DocumentNode, so that I call it document : D, but it's can be source.

PR #51, opened 2018-03-11
Add initial support for printing AST nodes

crisu83:
We still need to implement the printers for SDL later.

### DIFF
--- a/src/Language/Printer.php
+++ b/src/Language/Printer.php
@@ -2,17 +2,55 @@
 
 namespace Digia\GraphQL\Language;
 
+use Digia\GraphQL\Error\LanguageException;
 use Digia\GraphQL\Language\AST\Node\NodeInterface;
+use Digia\GraphQL\Language\Writer\WriterInterface;
 
 class Printer implements PrinterInterface
 {
+    /**
+     * @var array|WriterInterface[]
+     */
+    protected $writers;
+
+    /**
+     * Printer constructor.
+     * @param array|WriterInterface[] $writers
+     */
+    public function __construct(array $writers)
+    {
+        foreach ($writers as $writer) {
+            $writer->setPrinter($this);
+        }
+
+        $this->writers = $writers;
+    }
 
     /**
      * @inheritdoc
+     * @throws LanguageException
      */
     public function print(NodeInterface $node): string
     {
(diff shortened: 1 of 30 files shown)


crisu83:
Didn't add tests yet, but I think we can do that once the printer supports SDL. Added a separate issue for this #55.

src/Language/Writer/NullValueWriter.php
```diff
@@ -0,0 +1,26 @@
...
+use Digia\GraphQL\Language\AST\Node\NodeInterface;
+use Digia\GraphQL\Language\AST\Node\NullValueNode;
+
+class NullValueWriter extends AbstractWriter
+{
+    /**
+     * @param NodeInterface|NullValueNode $node
+     * @inheritdoc
+     */
+    public function write(NodeInterface $node): string
```
hungneox: If we return null here why we need the $node?

crisu83: It's the interface, it's required for all the other writers.

src/Language/Writer/IntValueWriter.php
```diff
@@ -0,0 +1,26 @@
...
+     */
+    public function write(NodeInterface $node): string
+    {
+        return $node->getValue();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function supportsWriter(NodeInterface $node): bool
```
hungneox: Should we consider prefix it with `Is`? `IsSupportsWriter`?

src/Language/Printer.php
```diff
@@ -2,17 +2,55 @@
...
+        }
+
+        $this->writers = $writers;
+    }
 
     /**
      * @inheritdoc
+     * @throws LanguageException
      */
     public function print(NodeInterface $node): string
```
hungneox: I guess NodeInterface is still a blank interface?

src/Language/Writer/IntValueWriter.php
crisu83: This is similar to other namings, I guess we can change it later.

src/Language/Printer.php
crisu83: No, there are some common methods.

PR #52, opened 2018-03-11
Visitor ancestors

crisu83:
Refactor visitors to support ancestors which is required to implement validation rules.


crisu83:
Rebased on master.

PR #54, opened 2018-03-11
ExecutableDefinitions validation rule

crisu83:


composer.json
```diff
@@ -20,6 +20,8 @@
     },
     "autoload": {
         "files": [
+            "./src/Error/formatError.php",
```
hungneox: I'm not sure, but it seems like we have too many autoloads file :)

src/Validation/Rule/ExecutableDefinitionRule.php
```diff
@@ -0,0 +1,44 @@
...
+ * operation or fragment definitions.
+ */
+class ExecutableDefinitionRule extends AbstractRule
+{
+    /**
+     * @inheritdoc
+     */
+    public function enterNode(NodeInterface $node): ?NodeInterface
+    {
+        if ($node instanceof DocumentNode) {
```
hungneox: I think it is easier to read if you return soon, return often :P

src/Validation/ValidationContext.php
```diff
@@ -0,0 +1,281 @@
...
+        $usages = $this->recursiveVariableUsages[(string)$operation] ?? null;
+
+        if (null === $usages) {
+            $usages    = $this->getVariableUsages($operation);
+            $fragments = $this->getRecursivelyReferencedFragments($operation);
+
+            foreach ($fragments as $fragment) {
+                // TODO: Figure out a more performance way to do this.
+                /** @noinspection SlowArrayOperationsInLoopInspection */
+                $usages = array_merge($usages, $this->getVariableUsages($fragment));
```
hungneox: Note that in order to merge 2 array deeply you need to use `array_merge_recursive`

src/Validation/Rule/ExecutableDefinitionRule.php
crisu83: I agree. I'll make this change.

src/Validation/ValidationContext.php
crisu83: This is a flat array.

src/Validation/Rule/ExecutableDefinitionRule.php
crisu83: Actually no, because we always want to return `$node`.

src/Validation/ValidationContext.php
hungneox: ok :D

PR #56, opened 2018-03-11
FieldOnCorrectType validation rule

crisu83:


Jalle19:
There are some namespace/autoloading issues that makes phpstan fail

crisu83:
@Jalle19 yes, I just fixed them.

src/Validation/Rule/FieldOnCorrectTypeRule.php
```diff
@@ -0,0 +1,137 @@
...
+ * parent type, or are an allowed meta field such as __typename.
+ */
+class FieldOnCorrectTypeRule extends AbstractRule
+{
+    /**
+     * @inheritdoc
+     */
+    public function enterNode(NodeInterface $node): ?NodeInterface
+    {
+        if ($node instanceof FieldNode) {
```
Jalle19: I would return early here to save one indent level

crisu83: Well yes, @hungneox said the same. However, if you look at the rest of the rules I think it's easier to understand if we leave it like this.

PR #58, opened 2018-03-11
KnownArgumentNames validation rule

crisu83:


Jalle19:
@crisu83 please rebase instead of merging master, it makes the diff much easier to read

crisu83:
@Jalle19 The error message for Syntax error is taken from the reference impl.

PR #64, opened 2018-03-11
LoneAnonymousOperation validation rule

crisu83:


src/Validation/Rule/LoneAnonymousOperationRule.php
```diff
@@ -0,0 +1,45 @@
...
+use Digia\GraphQL\Error\ValidationException;
+use Digia\GraphQL\Language\AST\Node\DocumentNode;
+use Digia\GraphQL\Language\AST\Node\NodeInterface;
+use Digia\GraphQL\Language\AST\Node\OperationDefinitionNode;
+use function Digia\GraphQL\Validation\anonymousOperationNotAloneMessage;
+
+/**
+ * Lone anonymous operation
+ *
+ * A GraphQL document is only valid if when it contains an anonymous operation
```
hugovk: Typo: "if when"

crisu83: Typo was taken directly from the reference impl. :D

PR #67, opened 2018-03-11
NoUnusedFragments validation rule

crisu83:


src/Validation/Rule/NoUnusedFragmentsRule.php
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * No unused fragments
+ *
+ * A GraphQL document is only valid if all fragment definitions are spread
+ * within operations, or spread within other fragments spread within operations.
+ */
+class NoUnusedFragmentsRule extends AbstractRule
+{
+    /**
+     * @var array|OperationDefinitionNode[]
```
Jalle19: `array` is superflous here

src/Validation/Rule/NoUnusedFragmentsRule.php
```diff
@@ -0,0 +1,75 @@
...
+     */
+    protected $operationDefinitions = [];
+
+    /**
+     * @var array|FragmentDefinitionNode[]
```
Jalle19: Same thing here

PR #68, opened 2018-03-11
NoUnusedVariables validation rule

crisu83:


src/Validation/Rule/NoUnusedVariablesRule.php
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * No unused variables
+ *
+ * A GraphQL operation is only valid if all variables defined by an operation
+ * are used, either directly or within a spread fragment.
+ */
+class NoUnusedVariablesRule extends AbstractRule
+{
+    /**
+     * @var array|VariableDefinitionNode[]
```
Jalle19: `array` is superflous here

src/Validation/Rule/NoUnusedVariablesRule.php
```diff
@@ -0,0 +1,73 @@
...
+     */
+    protected $variableDefinitions;
+
+    /**
+     * @inheritdoc
+     */
+    public function enterNode(NodeInterface $node): ?NodeInterface
+    {
+        if ($node instanceof OperationDefinitionNode) {
+            $this->variableDefinitions = [];
```
Jalle19: Is this correct? What if there are more than one operation definition nodes?

src/Validation/Rule/NoUnusedVariablesRule.php
```diff
@@ -0,0 +1,73 @@
...
+    public function leaveNode(NodeInterface $node): ?NodeInterface
+    {
+        if ($node instanceof OperationDefinitionNode) {
+            $variableNamesUsed = [];
+            $usages            = $this->validationContext->getRecursiveVariableUsages($node);
+            $operationNameNode = $node->getName();
+            $operationName     = null !== $operationNameNode ? $operationNameNode->getValue() : null;
+
+            /** @var VariableNode $variableNode */
+            foreach ($usages as ['node' => $variableNode]) {
```
Jalle19: This syntax looks weird, are you sure this is correct?

src/Validation/Rule/NoUnusedVariablesRule.php
crisu83: Yes, the variables are checked for each OperationDefinitionNode separately.

src/Validation/Rule/NoUnusedVariablesRule.php
crisu83: This is correct, it's the destructor syntax.

PR #69, opened 2018-03-11
OverlappingFieldsCanBeMerged validation rule

crisu83:


crisu83:
There is still one test case that fails, but I marked it as skipped and created a separate issue (#69) for it to be able to move on to the next rule.

hugovk:
> and created a separate issue (#69) for it 

Do you mean https://github.com/digiaonline/graphql-php/issues/76?

src/Validation/Conflict/ComparisonContext.php
```diff
@@ -0,0 +1,96 @@
...
+     */
+    protected $fieldMap = [];
+
+    /**
+     * @var array
+     */
+    protected $fragmentNames = [];
+
+    /**
+     * @var array|Conflict[]
```
Jalle19: `array` is superflous here

src/Validation/Conflict/ComparisonContext.php
```diff
@@ -0,0 +1,96 @@
...
+     */
+    public function hasConflicts(): bool
+    {
+        return !empty($this->conflicts);
+    }
+
+    /**
+     * @return array|Conflict[]
+     */
+    public function getConflicts()
```
Jalle19: Could have static type-hint

src/Validation/Conflict/FieldContext.php
```diff
@@ -0,0 +1,65 @@
...
+namespace Digia\GraphQL\Validation\Conflict;
+
+use Digia\GraphQL\Language\Node\FieldNode;
+use Digia\GraphQL\Type\Definition\CompositeTypeInterface;
+use Digia\GraphQL\Type\Definition\Field;
+
+class FieldContext
+{
+    /**
+     * @var CompositeTypeInterface|null
```
Jalle19: `?ComposerTypeInterface`

src/Validation/Conflict/FieldContext.php
```diff
@@ -0,0 +1,65 @@
...
+     */
+    protected $parentType;
+
+    /**
+     * @var FieldNode
+     */
+    protected $node;
+
+    /**
+     * @var Field|null
```
Jalle19: `?Field`

src/Validation/Conflict/FieldContext.php
```diff
@@ -0,0 +1,65 @@
...
+        FieldNode $node,
+        ?Field $definition = null
+    ) {
+        $this->parentType = $parentType;
+        $this->node       = $node;
+        $this->definition = $definition;
+    }
+
+    /**
+     * @return CompositeTypeInterface|null
```
Jalle19: Same for these as for the class properties

Jalle19:
@crisu83 would it perhaps be a good idea to use https://github.com/php-ds/polyfill instead of creating custom Map and Pair types?

src/Validation/Conflict/FieldContext.php
crisu83: I had it like this, but I got errors for it. Can't remember what complained about them.

crisu83:
@hugovk yes.

crisu83:
@Jalle19 I wouldn't introduce that dependency because we still have to implement the `PairSet` or something similar to be able to keep track of the fragments checked.

PR #83, opened 2018-03-12
Use built-in constants instead of custom ones

Jalle19:


Jalle19:
PHP_INT_MAX is larger on 64-bit systems than the old 2,147 billion people are used to

PR #90, opened 2018-03-12
ProvidedNonNullArguments validation rule

crisu83:


src/Provider/DirectivesProvider.php
```diff
@@ -39,7 +39,7 @@ public function register()
                     DirectiveLocationEnum::INLINE_FRAGMENT,
                 ],
                 'args'        => [
-                    'if ' => [
```
crisu83: Oops :D

PR #92, opened 2018-03-12
SingleFieldSubscription validation rule

crisu83:


src/Validation/Rule/SingleFieldSubscriptionsRule.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+class SingleFieldSubscriptionsRule extends AbstractRule
+{
+    /**
+     * @inheritdoc
+     */
+    public function enterNode(NodeInterface $node): ?NodeInterface
+    {
+        if ($node instanceof OperationDefinitionNode) {
+            if ($node->getOperation() === 'subscription') {
```
Jalle19: This could be combined with the outer if clause

PR #94, opened 2018-03-12
UniqueDirectivesPerLocation validation rule

crisu83:


src/Validation/Rule/UniqueDirectivesPerLocationRule.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+class UniqueDirectivesPerLocationRule extends AbstractRule
+{
+    /**
+     * @inheritdoc
+     */
+    public function enterNode(NodeInterface $node): ?NodeInterface
+    {
+        /** @var DirectiveNode[] $directives */
+        $directives = method_exists($node, 'getDirectives') ? $node->getDirectives() : null;
```
Jalle19: Is there no way to use instanceof instead here?

PR #96, opened 2018-03-13
Rules builder

### LINKED ISSUE #95: Validation Bug
# What did I do?
Use graphql() with validation for `MutationTest::testSimpleMutation`

```php
public function testSimpleMutation()
{
    $schema = GraphQLSchema([
        'mutation' =>
            new ObjectType([
                'name'   => 'M',
                'fields' => [
                    'greeting' => [
                        'type'    => GraphQLString(),
                        'resolve' => function ($source, $args, $context, $info) {
                            return sprintf('Hello %s.', $args['name']);
                        },
                        'args'    => [
                            'name' => [
                                'type' => GraphQLString()
                            ]
                        ]
                    ]
                ]
            ])
    ]);

    $source = '
    mutation M($name: String) {
        greeting(name:$name)
    }';

    $executionResult = graphql($schema, $source, '', null, ['name' => 'Han Solo']);

    $expected = new ExecutionResult([
        'greeting' => 'Hello Han Solo.'
    ], []);

    $this->assertEquals($expected, $executionResult);
}
```

## Run that single test case:
```
./vendor/bin/phpunit --filter testSimpleMutation
HPUnit 7.0.2 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.1.14 with Xdebug 2.6.0
Configuration: /Users/hunguyen/Codes/php/graphql-php/phpunit.xml

.                                                                   1 / 1 (100%)

Time: 184 ms, Memory: 10.00MB

OK (1 test, 1 assertion)
````

## Run the whole functional test

screen shot 2018-03-13 at 4 13 58

screen shot 2018-03-13 at 4 16 08

The weird thing is the error said something about `FragTwo` which is not include in the test.

```php
array_map(function(\Exception $ex) {
    printf("\n%s", $ex->getMessage());
},$executionResult->getErrors());
```

I got:
```
Fragment "FragOne" is never used.
Fragment "FragTwo" is never used
```

I suspect that there is something is cached inside the validation when run the whole testsuite.

crisu83:


src/Validation/Rule/SupportedRules.php
```diff
@@ -32,16 +27,15 @@ class SupportedRules
     /**
      * @return array
      */
-    public static function get(): array
+    public static function getNew(): array
```
Jalle19: Not sure if the method name really needs to change

crisu83: Maybe, I did it for clarity.

PR #97, opened 2018-03-13
UniqueFragmentNames validation rule

crisu83:


crisu83:
Merged early by mistake. All checks pass so it should be all fine.

PR #103, opened 2018-03-13
VariablesInAllowedPosition validation rule

crisu83:


src/Validation/Rule/VariablesInAllowedPositionRule.php
```diff
@@ -0,0 +1,119 @@
...
+     * @var TypeComparator
+     */
+    protected $typeComparator;
+
+    /**
+     * VariablesInAllowedPositionRule constructor.
+     */
+    public function __construct()
+    {
+        $this->typeComparator = new TypeComparator();
```
Jalle19: This is obviously not ideal but we can probably come up with a solution later.

crisu83: Yes, but this is the middle ground.

PR #104, opened 2018-03-13
Execute complete type value

hungneox:
+ Bunch of function to complete different types
+ Add one important test


src/Execution/ExecutionStrategy.php
```diff
@@ -85,20 +100,25 @@ protected function collectFields(
         $visitedFragmentNames
     ) {
         foreach ($selectionSet->getSelections() as $selection) {
-            /** @var FieldNode $selection */
             switch ($selection->getKind()) {
                 case NodeKindEnum::FIELD:
```
crisu83: Can't we use `instanceof` here instead of switch-case?

src/Execution/ExecutionStrategy.php
```diff
@@ -127,6 +147,60 @@ protected function collectFields(
...
 
+
+    /**
+     * @param $node
+     * @return bool
+     * @throws InvalidTypeException
+     * @throws \Digia\GraphQL\Error\ExecutionException
+     * @throws \Digia\GraphQL\Error\InvariantException
+     */
+    private function shouldIncludeNode(NodeInterface $node)
```
Jalle19: : bool

src/Execution/ExecutionStrategy.php
```diff
@@ -127,6 +147,60 @@ protected function collectFields(
...
+        return true;
+    }
+
+    /**
+     * @param FragmentDefinitionNode|InlineFragmentNode $fragment
+     * @param ObjectType                                $type
+     * @return bool
+     * @throws InvalidTypeException
+     */
+    private function doesFragmentConditionMatch(NodeInterface $fragment, ObjectType $type)
```
Jalle19: : bool

src/Execution/ExecutionStrategy.php
```diff
@@ -138,19 +212,20 @@ private function getFieldNameKey(FieldNode $node)
...
+     * @param ObjectType $objectType
+     * @param            $rootValue
      * @param            $path
      * @param            $fields
-     *
      * @return array
      * @throws InvalidTypeException
+     * @throws \Digia\GraphQL\Error\ExecutionException
+     * @throws \Digia\GraphQL\Error\InvariantException
+     * @throws \Throwable
```
Jalle19: Where does \Throwable come from?

src/Execution/ExecutionStrategy.php
```diff
@@ -284,58 +396,315 @@ private function buildResolveInfo(
...
+     */
+    private function completeValue(
+        TypeInterface $returnType,
+        $fieldNodes,
+        ResolveInfo $info,
+        $path,
+        &$result
+    ) {
+        if ($result instanceof \Exception || $result instanceof \Throwable) {
```
Jalle19: The first check is not needed, exceptions extend throwable

src/Execution/ExecutionStrategy.php
hungneox: screen shot 2018-03-15 at 1 06 15
It's deep down somewhere, PhpStorm suggested to add that tag.

PR #106, opened 2018-03-14
ValuesOnCorrectTypes validation rule

crisu83:


src/Validation/Rule/ValuesOfCorrectTypeRule.php
```diff
@@ -0,0 +1,275 @@
...
+class ValuesOfCorrectTypeRule extends AbstractRule
+{
+    /**
+     * @inheritdoc
+     */
+    protected function enterNullValue(NullValueNode $node): ?NullValueNode
+    {
+        $type = $this->validationContext->getInputType();
+
+        if ($type instanceof NonNullType) {
```
Jalle19: You could use the `isNonNullType` function you recently introduced

PR #108, opened 2018-03-14
Fix testMaintainsTypeInfoDuringEdit (fixes #42)

### LINKED ISSUE #42: Fix testMaintainsTypeInfoDuringEdit
One test is still failing from `VisitorTest`. The issue is most likely in the logic how we visit nodes added by visitors. I suspect that we are visited them in the wrong order.

crisu83:
Yay 🎉 Thanks @Jalle19.


crisu83:
Nevermind. this is probably not a fix, because it breaks another test case.

PR #111, opened 2018-03-15
Execution directives tests

hungneox:
* Porting all execution directives tests from `graphql-js`.
* Improve `collectFields` method


.idea/php-test-framework.xml
```diff
@@ -0,0 +1,14 @@
...
+      <tool tool_name="PHPUnit">
+        <cache>
+          <versions>
+            <info id="Local" version="7.0.2" />
+          </versions>
+        </cache>
+      </tool>
+    </tools_cache>
+  </component>
+</project>
```
Jalle19: I'm not 100% sure if this fime should be committed, it seems to contain configuration specific to your computer. Maybe we should ignore the file for now?

crisu83: I was actually thinking the same, but I guess it's ok.

PR #116, opened 2018-03-15
Resolver map

### LINKED ISSUE #114: Resolver map
We need to implement the support for giving a Resolver map to the `SchemaBuilder` to wire up schemas that are defined using SDL. Apollo's GraphQL Tools uses a similar approach (https://www.apollographql.com/docs/graphql-tools/resolvers.html#Resolver-map). However, we should implement this using classes instead of closures.

crisu83:
This pull request solves two issues:

- Basic type resolving (automatic)
- Setting resolvers on field definitions based on a resolver map

### DIFF
--- a/src/Language/LanguageProvider.php
+++ b/src/Language/LanguageProvider.php
@@ -2,8 +2,7 @@
 
 namespace Digia\GraphQL\Language;
 
-use Digia\GraphQL\Error\InvalidTypeException;
-use Digia\GraphQL\Language\Node\NamedTypeNode;
+use Digia\GraphQL\Execution\ValuesResolver;
 use Digia\GraphQL\Language\NodeBuilder\NodeBuilder;
 use Digia\GraphQL\Language\NodeBuilder\NodeBuilderInterface;
 use Digia\GraphQL\Language\NodeBuilder\SupportedBuilders;
@@ -50,12 +49,8 @@ class LanguageProvider extends AbstractServiceProvider
             ->withArgument(SupportedWriters::get());
 
         $this->container->add(DefinitionBuilderInterface::class, DefinitionBuilder::class, true/* $shared */)
-            ->withArguments([
-                function (NamedTypeNode $node) {
-                    throw new InvalidTypeException(sprintf('Type "%s" not found in document.', $node->getNameValue()));
-                },
-                CacheInterface::class,
-            ]);
+            ->withArgument(CacheInterface::class)
+            ->withArgument(ValuesResolver::class);
 
         $this->container->add(SchemaBuilderInterface::class, SchemaBuilder::class, true/* $shared */)
             ->withArgument(DefinitionBuilderInterface::class);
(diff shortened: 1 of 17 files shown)


tests/Functional/starWarsData.php
```diff
@@ -0,0 +1,135 @@
+<?php
+
+namespace Digia\GraphQL\Test\Functional;
+
+function luke()
```
hungneox: Just a trivial thing, but I think you can name it like `Luke()`, `Vader()` because they are proper noun

tests/Functional/QueryTest.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace Digia\GraphQL\Test;
+
+use function Digia\GraphQL\graphql;
+use function Digia\GraphQL\Test\Functional\starWarsSchema;
+
+class QueryTest extends TestCase
+{
+    public function testCorrectlyIdentifiesR2D2AsTheHeroOfTheStarWarsSaga()
```
hungneox: I think it is easier to read if you add a comment above this function
// Correctly identifies R2D as the hero of the star wars saga

tests/Functional/starWarsData.php
crisu83: This was taken from the reference implementation as is.

src/Language/LanguageProvider.php
```diff
@@ -13,6 +13,7 @@
 use Digia\GraphQL\Language\SchemaBuilder\SchemaBuilder;
 use Digia\GraphQL\Language\SchemaBuilder\SchemaBuilderInterface;
 use Digia\GraphQL\Language\Writer\SupportedWriters;
+use Digia\GraphQL\Type\SchemaInterface;
```
Jalle19: This looks unused?

src/Language/SchemaBuilder/DefinitionBuilderInterface.php
```diff
@@ -6,6 +6,7 @@
 use Digia\GraphQL\Language\Node\NodeInterface;
 use Digia\GraphQL\Type\Definition\DirectiveInterface;
 use Digia\GraphQL\Type\Definition\TypeInterface;
+use Digia\GraphQL\Type\SchemaInterface;
```
Jalle19: Doesn't seem to be needed?

crisu83: Fixed.

src/Language/LanguageProvider.php
crisu83: Fixed.

src/Language/SchemaBuilder/DefinitionBuilder.php
```diff
@@ -375,21 +408,43 @@ protected function getTypeDefinition(string $typeName): ?TypeDefinitionNodeInter
...
+    /**
+     * @param NodeInterface|EnumValueDefinitionNode|FieldDefinitionNode $node
+     * @return null|string
+     * @throws InvariantException
+     * @throws ExecutionException
+     * @throws InvalidTypeException
+     */
+    protected function getDeprecationReason(NodeInterface $node): ?string
+    {
+        $deprecated = (new ValuesResolver())->getDirectiveValues(GraphQLDeprecatedDirective(), $node);
```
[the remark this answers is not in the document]
crisu83: Fixed.

PR #117, opened 2018-03-15
Execution abstract first test

hungneox:
* Change \ArrayObject to array for easier to debug
* First Abstract test

### DIFF
--- a/src/Execution/ExecutionStrategy.php
+++ b/src/Execution/ExecutionStrategy.php
@@ -100,8 +100,8 @@ abstract class ExecutionStrategy
     protected function collectFields(
         ObjectType $runtimeType,
         SelectionSetNode $selectionSet,
-        $fields,
-        $visitedFragmentNames
+        &$fields,
+        &$visitedFragmentNames
     ) {
         foreach ($selectionSet->getSelections() as $selection) {
             // Check if this Node should be included first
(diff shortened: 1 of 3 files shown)


src/Execution/ExecutionStrategy.php
```diff
@@ -189,7 +189,7 @@ private function doesFragmentConditionMatch(
 
         $conditionalType = typeFromAST($this->context->getSchema(), $typeConditionNode);
 
-        if ($conditionalType === $type) {
+        if (gettype($conditionalType) === gettype($type)) {
```
crisu83: I'm not a fan of this solution. Simply comparing the instances should work.

src/Execution/ExecutionStrategy.php
```diff
@@ -767,12 +776,12 @@ private function collectAndExecuteSubFields(
...
                     $fieldNode->getSelectionSet(),
                     $subFields,
-                    new \ArrayObject()
+                    $visitedFragmentNames
                 );
             }
         }
 
-        if ($subFields->count()) {
+        if (count($subFields) > 0) {
```
crisu83: Can't we use `!empty($subFields)`?

src/Execution/ExecutionStrategy.php
hungneox: The issue is It didn't work :<

src/Execution/ExecutionStrategy.php
hungneox: `if(count($subFields))` also works. But this for clarity.

PR #119, opened 2018-03-16
Refactoring Abstract tests

hungneox:
* Figured out why `===` doesn't work
* Amend `ResolveTypeTrait` to return null when there is no `resolveTypeFunction`


src/Type/Definition/ResolveTypeTrait.php
```diff
@@ -12,11 +12,15 @@
...
     /**
      * @param array ...$args
-     * @return TypeInterface|null
+     * @return TypeInterface|string|null
      */
-    public function resolveType(...$args): ?TypeInterface
+    public function resolveType(...$args)
     {
-        return \call_user_func_array($this->resolveTypeFunction, $args);
+        if(!empty($this->resolveTypeFunction)) {
```
hungneox: `graphql-js` library also support "resolveType allows resolving with type name"

crisu83: Use `isset`.

tests/Functional/Execution/AbstractTest.php
```diff
@@ -14,31 +13,161 @@
...
     /**
      * isTypeOf used to resolve runtime type for Interface
+     *
      * @throws \Digia\GraphQL\Error\InvariantException
      * @throws \Digia\GraphQL\Error\SyntaxErrorException
      */
-    public function testResolveFunctionForInterface()
+    public function testIsTypeOfUsedToResolveFunctionForInterface()
     {
+        $PetInterfaceType = GraphQLInterfaceType([
```
crisu83: I would use class fields for these.

tests/Functional/Execution/AbstractTest.php
```diff
@@ -14,31 +13,161 @@
...
             'query' => GraphQLObjectType([
                 'name'   => 'Query',
                 'fields' => [
                     'pets' => [
-                        'type'    => GraphQLList(PetType()),
+                        'type'    => GraphQLList($PetInterfaceType),
                         'resolve' => function ($source, $args, $context, $info) {
-                            return [new Dog('Odie', true), new Cat('Garfield', false)];
+                            return [
+                                new Dog('Odie', true),
```
crisu83: Remember to add this file to the `autoload-dev` `files`.

tests/Functional/Execution/AbstractTest.php
```diff
@@ -71,6 +200,327 @@ public function testResolveFunctionForInterface()
...
+     * @throws \Digia\GraphQL\Error\InvariantException
+     * @throws \Digia\GraphQL\Error\SyntaxErrorException
+     */
+    public function testResolveTypeOnInterfaceYieldsUsefulError()
+    {
+        $PetInterfaceType = GraphQLInterfaceType([
+            'name'        => 'Pet',
+            'resolveType' => function ($result, $context, $info) use (&$DogType, &$CatType, &$HumanType) {
+                if ($result instanceof Dog) {
+                    return $DogType;
```
crisu83: The Apollo implementation supports just returning the name of the type. This would probably be a better approach to ensure that you don't have multiple instances of the same types.

tests/Functional/Execution/AbstractTest.php
```diff
@@ -79,7 +529,7 @@ class Human
 
     public function __construct(string $name)
     {
-        $this->name;
+        $this->name = $name;
```
crisu83: Had to fix these in my PR too.

tests/Functional/Execution/AbstractTest.php
```diff
@@ -14,31 +13,161 @@
...
+    public function testIsTypeOfUsedToResolveFunctionForInterface()
     {
+        $PetInterfaceType = GraphQLInterfaceType([
+            'name'   => 'Pet',
+            'fields' => [
+                'name' => ['type' => GraphQLString()]
+            ]
+        ]);
+
+        $DogType = GraphQLObjectType([
```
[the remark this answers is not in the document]
hungneox: Not really the same object, basically we can't use Dog() for this, because it will create two identical objects in the schema. That's yesterday issue with `===`

tests/Functional/Execution/AbstractTest.php
hungneox: why?

tests/Functional/Execution/AbstractTest.php
hungneox: What do you mean? : D

tests/Functional/Execution/AbstractTest.php
hungneox: Please check the test below `resolveType allows resolving with type name`

PR #120, opened 2018-03-16
Rules provider

crisu83:


tests/Functional/Execution/AbstractTest.php
```diff
@@ -98,7 +98,7 @@ public function __construct(string $name, bool $woofs)
 class Cat
 {
     public $name;
-    public $woofs;
+    public $meows;
```
hungneox: I fixed these twos minors in other PR :">

PR #121, opened 2018-03-16
Remove unused classes - re-add later if needed

hungneox:
* Remove unused classes - re-add later if needed


src/Execution/Resolver/PropertyFieldResolver.php
```diff
@@ -1,36 +0,0 @@
-<?php
-
-namespace Digia\GraphQL\Execution\Resolver;
-
-use Digia\GraphQL\Error\ExecutionException;
-use Digia\GraphQL\Execution\ExecutionEnvironment;
-
-class PropertyFieldResolver implements ResolverInterface
```
crisu83: I would like to keep this class based if possible because we will need to add resolver classes soon.

crisu83: You could wire this up through DI instead of using a method.

PR #129, opened 2018-03-17
Refactor valueFromAST

crisu83:


src/Execution/ValuesResolver.php
```diff
@@ -75,49 +91,76 @@ public function coerceArgumentValues($objectType, $field, array $variableValues
...
-                    $argumentType,
-                    $argumentName,
-                    $variableValues
-                );
+                $coercedValue = null;
+
+                try {
+                    $coercedValue = $this->coerceValueFromAST($argumentNode->getValue(), $argumentType,
+                        $variableValues);
+                } catch (CoercingException $ex) {
```
hungneox: I just wonder should we move the logic in this `else` block to `coerceValueFromAST` method?

src/Execution/ExecutionStrategy.php
```diff
@@ -79,7 +80,8 @@ public function __construct(
         $this->context        = $context;
         $this->operation      = $operation;
         $this->rootValue      = $rootValue;
-        $this->valuesResolver = new ValuesResolver();
+        // TODO: Inject the ValuesResolver instance by using a builder for this class.
+        $this->valuesResolver = new ValuesResolver(new ValueNodeCoercer());
```
hungneox: @crisu83 actually I think we can also inject $valueResolver from the container :D

src/Execution/ValuesResolver.php
crisu83: The reason I changed this is because that method is used elsewhere and there we’re not coercing args.

src/Execution/ExecutionStrategy.php
crisu83: That was the point.

crisu83: That was the point.

PR #130, opened 2018-03-17
Execution union interface test

hungneox:
* Execution Uniion interface tests
* Fix some issue in  `introspection` definition

### DIFF
--- a/src/Execution/ExecutionStrategy.php
+++ b/src/Execution/ExecutionStrategy.php
@@ -349,6 +349,7 @@ abstract class ExecutionStrategy
     ) {
         /** @var FieldNode $fieldNode */
         $fieldNode = $fieldNodes[0];
+        $fieldName = $fieldNode->getNameValue();
 
         $field = $this->getFieldDefinition($this->context->getSchema(), $parentType, $fieldNode->getNameValue());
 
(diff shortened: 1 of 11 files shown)


src/Execution/ExecutionStrategy.php
```diff
@@ -241,6 +241,11 @@ protected function executeFields(
                 $fieldPath
             );
 
+            // does not include illegal fields in output
+            if ($result === self::UNDEFINED) {
```
crisu83: Don’t do this. Use exceptions instead to indicate failure.

src/Execution/ExecutionStrategy.php
```diff
@@ -339,11 +349,12 @@ protected function resolveField(
...
         /** @var FieldNode $fieldNode */
         $fieldNode = $fieldNodes[0];
+        $fieldName = $fieldNode->getNameValue();
 
         $field = $this->getFieldDefinition($this->context->getSchema(), $parentType, $fieldNode->getNameValue());
 
         if (!$field) {
-            return null;
+            return self::UNDEFINED;
         }
```
crisu83: Throw appropriate exception instead.

PR #131, opened 2018-03-18
Execution abstract promise test

hungneox:
* Use ReactPHP Promise
* Port first two tests from `abstract-promise-test.js`
* Handles promise returns from`resolve` and `isTypeOf`

### DIFF
--- a/src/Execution/ExecutionStrategy.php
+++ b/src/Execution/ExecutionStrategy.php
@@ -23,11 +23,13 @@ use Digia\GraphQL\Type\Definition\ObjectType;
 use Digia\GraphQL\Type\Definition\TypeInterface;
 use Digia\GraphQL\Type\Definition\UnionType;
 use Digia\GraphQL\Type\Schema;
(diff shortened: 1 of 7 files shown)


src/Execution/ExecutionStrategy.php
```diff
@@ -238,9 +241,23 @@ protected function executeFields(
...
+                $isContainsPromise = true;
+            }
+
             $finalResults[$fieldName] = $result;
         }
 
+        if ($isContainsPromise) {
+            $keys    = array_keys($finalResults);
+            $promise = \React\Promise\all(array_values($finalResults));
+            $promise->then(function ($values) use ($keys, &$finalResults) {
```
hungneox: @crisu83 what do you think about passing `$finalResults` as ref?

crisu83: I think it's acceptable in this case, however, it might be a little bit slower. I would use `array_reduce` instead.

PR #135, opened 2018-03-19
Fix QueryTest

hungneox:
- Fix startWar query test

### DIFF
--- a/src/Execution/ExecutionStrategy.php
+++ b/src/Execution/ExecutionStrategy.php
@@ -23,6 +23,7 @@ use Digia\GraphQL\Type\Definition\ObjectType;
 use Digia\GraphQL\Type\Definition\TypeInterface;
 use Digia\GraphQL\Type\Definition\UnionType;
 use Digia\GraphQL\Type\Schema;
+use Digia\GraphQL\Util\ValueNodeCoercer;
 use React\Promise\ExtendedPromiseInterface;
 use React\Promise\PromiseInterface;
 use function Digia\GraphQL\Type\SchemaMetaFieldDefinition;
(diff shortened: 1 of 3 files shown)


src/Execution/ExecutionStrategy.php
```diff
@@ -732,12 +732,13 @@ private function defaultTypeResolver($value, $context, ResolveInfo $info, Abstra
 
         foreach ($possibleTypes as $index => $type) {
             $isTypeOfResult = $type->isTypeOf($value, $context, $info);
-
             if (null !== $isTypeOfResult) {
                 if ($this->isPromise($isTypeOfResult)) {
                     $promisedIsTypeOfResults[$index] = $isTypeOfResult;
                 } elseif ($isTypeOfResult) {
                     return $type;
+                } elseif(is_array($value) && $value['type'] === $type->getName()) { //@TODO Make `type` configurable
```
crisu83: `} elseif (\is_array($value) && isset($value['type']) && $value['type'] === $type->getName()) {`

tests/Functional/QueryTest.php
```diff
@@ -20,7 +20,6 @@ public function testCorrectlyIdentifiesR2D2AsTheHeroOfTheStarWarsSaga()
         /** @noinspection PhpUnhandledExceptionInspection */
         $result = graphql(starWarsSchema(), $query);
 
-        $this->markTestIncomplete('INCOMPLETE: We cannot yet resolve data for interface types.');
```
crisu83: @hungneox Thanks, for fixing this test case. I'll continue porting the rest of the query tests soon.

PR #139, opened 2018-03-19
Error location and path

### LINKED ISSUE #138: Error location and path are missing
Currently, we cannot get the location nor the path from errors that happen inside resolvers. We need this in order to present the error as a whole to the user. I had to mark the three last test cases in the `QueryTest` as incomplete because of this missing feature (#137).

crisu83:
Fixes #138


src/Error/helpers.php
```diff
@@ -51,18 +72,18 @@ function highlightSourceAtLocation(Source $source, SourceLocation $location): st
...
     $lineNum        = (string)$contextLine;
     $nextLineNum    = (string)($contextLine + 1);
-    $padLen         = mb_strlen($nextLineNum);
-    $lines          = preg_split("/\r\n|[\n\r]/", $source->getBody());
+    $padLen         = \mb_strlen($nextLineNum);
+    $lines          = \preg_split("/\r\n|[\n\r]/", $source->getBody());
     $lines[0]       = whitespace($locationOffset->getColumn() - 1) . $lines[0];
-    $outputLines = [
-        sprintf('%s (%s:%s)', $source->getName(), $contextLine, $contextColumn),
+    $outputLines    = [
```
hungneox: This block looks like Perl code :P

PR #142, opened 2018-03-20
Execution mutation test

hungneox:
* Implement `executeFieldsSerially`, I will refactor it later
* Porting tests from `mutation-test.js`


src/Execution/ExecutionStrategy.php
```diff
@@ -305,6 +326,7 @@ public function executeFieldsSerially(
      * @param string     $fieldName
      * @return \Digia\GraphQL\Type\Definition\Field|null
```
crisu83: Should we import these?

PR #144, opened 2018-03-20
Execution list tests

hungneox:

### DIFF
--- a/tests/Functional/Execution/ListTest.php
+++ b/tests/Functional/Execution/ListTest.php
@@ -1,0 +1,131 @@
+<?php
+
+namespace Digia\GraphQL\Test\Functional\Execution;
+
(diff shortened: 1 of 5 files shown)


tests/Functional/Execution/ListTest.php
```diff
@@ -0,0 +1,347 @@
...
+    // EXECUTE: ACCEPTS ANY ITERABLE AS LIST VALUE
+
+    /**
+     * @param $testType
+     * @param $testData
+     * @param $expected
+     * @throws \Digia\GraphQL\Error\InvariantException
+     * @throws \Digia\GraphQL\Error\SyntaxErrorException
+     */
+    public function makeTest($testType, $testData, $expected)
```
crisu83: I would make this `protected` and think of a better name for it.

PR #145, opened 2018-03-20
Refactor parser

crisu83:
- Change suffix for node builders to `NodeBuilder` (from `Builder`) to avoid confusion
- Change name of `NodeBuilder` (actually builder builder) to `NodeDirector`
- Introduce `ASTDirector` for building the AST using the builder pattern

This will make it easier for us to maintain the parser in the long run as well as add new features if the specification gets expanded later.


crisu83:
This PR is pretty hard to review because of the number of changes (sorry about that). I'll just go ahead and merge this because all test seem to be passing.

PR #155, opened 2018-03-25
Refactor public API

crisu83:
Refactor the public API to make it easier for developers to use the library, even without using the namespaced functions found in [api.php](../tree/master/src/api.php). I also changed the `graphql` function to return an array instead of `ExecutionResult`. This makes testing a bit more simple because we do not have to instantiate new `ExecutionResult` objects to be able to assert results.

### DIFF
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -115,7 +115,7 @@ class GraphQL
      * @param array  $args
      * @return mixed
      */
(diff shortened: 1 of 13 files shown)


crisu83:
@hungneox I would even consider removing all the functions except for `graphql` from `api.php`.

PR #158, opened 2018-03-29
Refactor helper functions to use the container

crisu83:
This will simplify usage of helper functions throughout the library, while still allowing the developer to inject their own implementations.

The way this is implemented is the following:

- Helper classes that are injected into the container
- Namespaced functions that expose methods on these helpers
- Direct usage of these functions across the project


composer.json
```diff
@@ -26,14 +26,15 @@
     "autoload": {
         "files": [
             "./src/Error/helpers.php",
+            "./src/Execution/values.php",
             "./src/Language/blockStringValue.php",
-            "./src/Language/helpers.php",
+            "./src/Language/utils.php",
```
hungneox: I would a void general name like `utils`. Because we have two `utils.php` file here. Even `/Util/util`.php

src/Util/utils.php
```diff
@@ -0,0 +1,201 @@
...
+        $map[$keyFn($item)] = $valFn($item);
+        return $map;
+    }, []);
+}
+
+/**
+ * @param $value
+ * @return string
+ */
+function toString($value): string
```
hungneox: I just wonder can we rename it to `getType` or something like that? `toString` doesn't reflect its function

composer.json
hungneox: Could we have `/utils/language.php` , `/utils/type.php` for example?

tests/Functional/Execution/ValuesResolverTest.php
```diff
@@ -58,7 +49,7 @@ public function testGetValues()
             $schema, [], null, null, ['name' => 'Han Solo'], null, $operation, []
         );
 
-        $args = $this->valuesResolver->coerceArgumentValues($definition, $node, $context->getVariableValues());
+        $args = getArgumentValues($definition, $node, $context->getVariableValues());
```
hungneox: I just think `getArgumentValues` is not the right name for the function, because values is really got from `$context->getVariablesValues`

composer.json
crisu83: What would be a better name?

src/Util/utils.php
crisu83: It does, it works for any kind of value; objects, array, scalars, null, etc.

tests/Functional/Execution/ValuesResolverTest.php
crisu83: That's what they call it in the reference implementation so I'd prefer to keep it like that for now.

src/Util/utils.php
hungneox: But normally, I expects to get a string value of given variable, instead of its type as string.

tests/Functional/Execution/ValuesResolverTest.php
hungneox: Actually I commented on this before, in the graphql specification they call it `coerceArgumentValues`

composer.json
hungneox: @crisu83 see my comment above : D

src/Execution/values.php
```diff
@@ -0,0 +1,44 @@
...
+use Digia\GraphQL\Type\Definition\Field;
+use Digia\GraphQL\Type\SchemaInterface;
+
+/**
+ * @param SchemaInterface          $schema
+ * @param VariableDefinitionNode[] $nodes
+ * @param array                    $inputs
+ * @return array
+ */
+//function getVariableValues(SchemaInterface $schema, array $nodes, array $inputs): array
```
hungneox: Why you comment this out?

tests/Functional/Execution/ValuesResolverTest.php
crisu83: I can change this, should I also change the associated directive-method?

src/Execution/values.php
crisu83: There is no such method in `ValuesHelper`.

composer.json
crisu83: We could, but on the other hand I would like the packages to be self-contained, if possible.

PR #180, opened 2018-04-04
Schema extension resolver registry

crisu83:
Add support for providing a resolver registry when extending a schema and add schema extension to the public API (had forgot about this earlier).


src/GraphQL.php
```diff
@@ -144,6 +145,31 @@ public static function buildSchema($source, $resolverRegistry, array $options =
             );
     }
 
+    /**
+     * @param SchemaInterface                 $schema
+     * @param string|Source                   $source
+     * @param array|ResolverRegistryInterface $resolverRegistry
```
Jalle19: Should this really be `array` or should it be `callable`?

crisu83: It is correct, that's if you want to provide the resolver registry as an array (it will be instantiated as a real `ResolverRegistry` inside the `GraphQL` class.

PR #183, opened 2018-04-05
Missing schema validation tests

crisu83:
Add missing schema validation tests that were initially skipped because we did not support schema extension back when they were written.

### DIFF
--- a/src/Schema/DefinitionBuilder.php
+++ b/src/Schema/DefinitionBuilder.php
@@ -246,9 +246,12 @@ class DefinitionBuilder implements DefinitionBuilderInterface
         return newObjectType([
             'name'        => $node->getNameValue(),
             'description' => $node->getDescriptionValue(),
-            'fields'      => function () use ($node) {
+            'fields'      => $node->hasFields() ? function () use ($node) {
                 return $this->buildFields($node);
-            },
+            } : [],
+            // Note: While this could make early assertions to get the correctly
+            // typed values, that would throw immediately while type system
+            // validation with validateSchema() will produce more actionable results.
             'interfaces'  => function () use ($node) {
                 return $node->hasInterfaces() ? \array_map(function (NodeInterface $interface) {
                     return $this->buildType($interface);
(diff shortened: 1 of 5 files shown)


src/Schema/DefinitionBuilder.php
```diff
@@ -246,9 +246,12 @@ protected function buildObjectType(ObjectTypeDefinitionNode $node): ObjectType
         return newObjectType([
             'name'        => $node->getNameValue(),
             'description' => $node->getDescriptionValue(),
-            'fields'      => function () use ($node) {
+            'fields'      => $node->hasFields() ? function () use ($node) {
```
crisu83: I noticed an inconsistency in how fields were handled compared to e.g. interfaces, so I decided to change this.

src/Schema/Extension/ExtendInfo.php
```diff
@@ -73,7 +73,7 @@ public function hasTypeExtensions(string $typeName): bool
      */
     public function getTypeExtensions(string $typeName): ?array
     {
-        return $this->typeExtensionsMap[$typeName] ?? null;
+        return $this->typeExtensionsMap[$typeName] ?? [];
```
crisu83: Extension AST Nodes are of type `array` so returning null was clearly a bug.

src/Schema/Extension/ExtensionContext.php
```diff
@@ -264,14 +264,14 @@ protected function extendInterfaceType(InterfaceType $type): InterfaceType
...
-            'name'           => $typeName,
-            'description'    => $type->getDescription(),
-            'fields'         => function () use ($type) {
+            'name'              => $typeName,
+            'description'       => $type->getDescription(),
+            'fields'            => function () use ($type) {
                 return $this->extendFieldMap($type);
             },
-            'astNode'        => $type->getAstNode(),
-            'typeExtensions' => $extensionASTNodes,
```
crisu83: This bug took me an hour to find. Better be when refactoring in the future!

crisu83:
@Jalle19 we can do that later, there is no such test in the reference implementation and in order to get this ready as soon as possible I've left all the additional test cases until later.

PR #184, opened 2018-04-05
Lexer unit tests

### LINKED ISSUE #9: Lexer unit tests
We need to port the test cases from the reference implementation found here:
https://github.com/graphql/graphql-js/blob/master/src/language/__tests__/lexer-test.js

You can look at our other test cases if you want to see an example.

### LINKED ISSUE #182: Fix lexer multi-byte test case failure

Jalle19:
Fixes #182, closes #9

### DIFF
--- a/src/Language/ASTBuilder.php
+++ b/src/Language/ASTBuilder.php
@@ -987,20 +987,11 @@ class ASTBuilder implements ASTBuilderInterface
         $token = $this->expect($lexer, TokenKindEnum::NAME);
         $value = $token->getValue();
 
-        if ($this->isOperation($value)) {
+        if (isOperation($value)) {
             return $value;
         }
 
         throw $this->unexpected($lexer, $token);
-    }
-
-    /**
-     * @param string $value
-     * @return bool
-     */
-    protected function isOperation(string $value): bool
-    {
-        return \in_array($value, ['query', 'mutation', 'subscription'], true);
     }
 
     /**
--- a/src/Language/Lexer.php
+++ b/src/Language/Lexer.php
@@ -63,6 +63,7 @@ class Lexer implements LexerInterface
         $startOfFileToken = new Token(TokenKindEnum::SOF);
 
         $reader->setLexer($this);
+
         $this->reader    = $reader;
         $this->lastToken = $startOfFileToken;
         $this->token     = $startOfFileToken;
@@ -177,24 +178,6 @@ class Lexer implements LexerInterface
     }
 
     /**
-     * @param int   $code
-     * @param int   $pos
-     * @param int   $line
-     * @param int   $col
-     * @param Token $prev
-     * @return Token
-     * @throws SyntaxErrorException
-     */
-    public function read(int $code, int $pos, int $line, int $col, Token $prev): Token
-    {
-        if ($token = $this->reader->read($code, $pos, $line, $col, $prev)) {
-            return $token;
-        }
-
-        throw new SyntaxErrorException($this->source, $pos, $this->unexpectedCharacterMessage($code));
-    }
-
-    /**
      * @param Token $prev
      * @return Token
      * @throws SyntaxErrorException
@@ -202,7 +185,7 @@ class Lexer implements LexerInterface
     protected function readToken(Token $prev): Token
     {
         $body       = $this->source->getBody();
-        $bodyLength = mb_strlen($body);
+        $bodyLength = \mb_strlen($body);
 
         $pos  = $this->positionAfterWhitespace($body, $prev->getEnd());
         $line = $this->line;
@@ -214,29 +197,33 @@ class Lexer implements LexerInterface
 
         $code = charCodeAt($body, $pos);
 
-        if (isSourceCharacter($code)) {
-            throw new SyntaxErrorException(
-                $this->source,
-                $pos,
-                sprintf('Cannot contain the invalid character %s', printCharCode($code))
-            );
+        $token = $this->reader->read($body, $bodyLength, $code, $pos, $line, $col, $prev);
+
+        if (null !== $token) {
+            return $token;
         }
 
-        return $this->read($code, $pos, $line, $col, $prev);
+        throw new SyntaxErrorException($this->source, $pos, $this->unexpectedCharacterMessage($code));
     }
 
     /**
+     * Report a message that an unexpected character was encountered.
+     *
      * @param int $code
      * @return string
      */
     protected function unexpectedCharacterMessage(int $code): string
     {
+        if (isSourceCharacter($code) && !isLineTerminator($code)) {
+            return \sprintf('Cannot contain the invalid character %s.', printCharCode($code));
+        }
+
         if ($code === 39) {
             // '
             return 'Unexpected single quote character (\'), did you mean to use a double quote (")?';
         }
 
-        return sprintf('Cannot parse the unexpected character %s', printCharCode($code));
+        return \sprintf('Cannot parse the unexpected character %s.', printCharCode($code));
     }
 
     /**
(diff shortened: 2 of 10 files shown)


src/Language/Reader/BlockStringReader.php
```diff
@@ -85,6 +85,10 @@ public function supportsReader(int $code, int $pos): bool
      */
     protected function isTripleQuote(string $body, int $code, int $pos): bool
     {
+        if (mb_strlen($body) < 3) {
```
crisu83: (nit) `\mb_string`

src/Language/Reader/StringReader.php
```diff
@@ -52,7 +52,7 @@ public function read(int $code, int $pos, int $line, int $col, Token $prev): Tok
 
             if ($code === 92) {
                 // \
-                $value .= sliceString($body, $chunkStart, $pos + 1);
```
crisu83: Good catch.

src/Language/Reader/StringReader.php
```diff
@@ -114,7 +115,7 @@ public function read(int $code, int $pos, int $line, int $col, Token $prev): Tok
             }
         }
 
-        throw new SyntaxErrorException($this->lexer->getSource(), $pos, 'Unterminated string.');
+        throw new SyntaxErrorException($this->lexer->getSource(), $pos - 1, 'Unterminated string.');
```
crisu83: Is this really a correct fix?

src/Language/Reader/StringReader.php
```diff
@@ -124,7 +125,15 @@ public function supportsReader(int $code, int $pos): bool
     {
         $body = $this->lexer->getBody();
 
-        return $code === 34 && charCodeAt($body, $pos + 1) !== 34;
+        if ($code === 34) {
```
crisu83: Please use early return instead.

tests/Unit/Language/LexerTest.php
```diff
@@ -0,0 +1,189 @@
...
+        $this->assertLexerTokenPropertiesEqual('"""' . "unescaped \\n\\r\\b\\t\\f\\u1234" . '""""',
+            TokenKindEnum::BLOCK_STRING,
+            0, 32,
+            'unescaped \\n\\r\\b\\t\\f\\u1234');
+        $this->assertLexerTokenPropertiesEqual('"""' . "slashes \\\\ \\/" . '""""',
+            TokenKindEnum::BLOCK_STRING,
+            0, 19,
+            "slashes \\\\ \\/");
+
+        $multiline = <<<EOD
```
crisu83: `EOD`?

crisu83:
Rebased on performance improvements by @fubhy and added missing test cases.

**NOTE:** Unicode-related tests are still failing, but I created a separate issue for them (#192).

src/Language/ASTBuilder.php
```diff
@@ -987,22 +987,13 @@ protected function parseOperationType(LexerInterface $lexer): string
...
         }
 
         throw $this->unexpected($lexer, $token);
     }
 
-    /**
-     * @param string $value
-     * @return bool
-     */
-    protected function isOperation(string $value): bool
```
crisu83: Change to use the function from utils.

src/Language/Lexer.php
```diff
@@ -176,24 +177,6 @@ public function setOptions(array $options)
...
-    /**
-     * @param int   $code
-     * @param int   $pos
-     * @param int   $line
-     * @param int   $col
-     * @param Token $prev
-     * @return Token
-     * @throws SyntaxErrorException
-     */
-    public function read(int $code, int $pos, int $line, int $col, Token $prev): Token
```
crisu83: This method felt unnecessary after @fubhy improvements.

src/Language/TokenReaderInterface.php
```diff
@@ -6,20 +6,16 @@
...
+     * @return Token|null
      */
-    public function read(
-      int $code,
-      int $pos,
-      int $line,
-      int $col,
-      Token $prev
-    ): Token;
+    public function read(string $body, int $bodyLength, int $code, int $pos, int $line, int $col, Token $prev): ?Token;
```
crisu83: Change the signature to accept the `$body` and `$bodyLength` so that we can void calculating the body length multiple times (this can be an expensive operation as we are using `mb_strlen` to do so).

src/Language/utils.php
```diff
@@ -17,19 +17,21 @@ function chrUTF8(int $cp)
...
 }
 
 /**
  * @param string $string
  * @param int    $position
+ * @param bool   $convertEncoding
  * @return int
  */
-function charCodeAt(string $string, int $position): int
+function charCodeAt(string $string, int $position, bool $convertEncoding = false): int
```
crisu83: Added a flag to skip converting the encoding in cases where it's not required. This is not the best solution but I think it's acceptable in this case.

src/Language/utils.php
```diff
@@ -38,14 +40,14 @@ function charCodeAt(string $string, int $position): int
  */
 function printCharCode(int $code): string
 {
-    if ($code === null) {
+    if ($code === 0x0000) {
```
crisu83: This was actually a bug. It appears that `0` equals `<EOF>` in PHP.

src/Language/utils.php
```diff
@@ -75,8 +77,7 @@ function isLetter(int $code): bool
  */
 function isNumber(int $code): bool
 {
-    // - or 0-9
-    return $code === 45 || ($code >= 48 && $code <= 57);
+    return $code >= 48 && $code <= 57; // 0-9
```
crisu83: This was a bug which caused the lexer to accept names with `dashes` in them (which it shouldn't).

src/Language/TokenReader.php
```diff
@@ -644,109 +520,9 @@ protected function readDigits(int $code, int $pos): int
...
         );
     }
-
-    /**
-     * @TODO Move to utils.
-     *
-     * @param int $code
-     * @return bool
-     */
-    protected function isLineTerminator(int $code): bool
```
crisu83: Move these methods to utils.

src/Language/utils.php
```diff
@@ -3,23 +3,53 @@
...
+ * Multi-byte compatible `ord`.
+ *
+ * Based on the Symfony's Mbstring polyfills.
+ *
  * @param string $string
+ * @param string $encoding
  * @return int
  */
-function ordUTF8(string $string)
+function mbOrd(string $s)
```
crisu83: Decided to port the `chr` and `ord` functions from Symfony's Mbstring polyfill.

src/Language/SourceLocation.php
```diff
@@ -57,11 +57,12 @@ public static function fromSource(Source $source, int $position): self
...
         $column  = $position + 1;
         $matches = [];
-        preg_match_all("/\r\n|[\n\r]/", mb_substr($source->getBody(), 0, $position), $matches, PREG_OFFSET_CAPTURE);
+
+        \preg_match_all("/\r\n|[\n\r]/", \substr($source->getBody(), 0, $position), $matches, PREG_OFFSET_CAPTURE);
 
         foreach ($matches[0] as $index => $match) {
             $line   += 1;
-            $column = $position + 1 - ($match[1] + mb_strlen($match[0]));
+            $column = $position + 1 - ($match[1] + \strlen($match[0]));
```
fubhy: Why can we use strlen here but have to use mb_strlen otherwise? Can't we use strlen() everywhere considering that the graphql spec doesn't allow multibyte characters anyways afaik?

src/Language/TokenReaderInterface.php
fubhy: Very good. :)

src/Language/utils.php
fubhy: Great. That should help a lot.

src/Language/utils.php
```diff
@@ -3,23 +3,53 @@
...
 {
-    [, $ord] = unpack('N', mb_convert_encoding($string, 'UCS-4BE', 'UTF-8'));
+    /** @noinspection CallableParameterUseCaseInTypeContextInspection */
+    $code = ($s = \unpack('C*', \substr($s, 0, 4))) ? $s[1] : 0;
```
fubhy: Same here. Why is it okay to use substr() here (vs mb_substr()).

src/Language/utils.php
```diff
@@ -128,8 +205,7 @@ function uniCharCode(string $a, string $b, string $c, string $d): string
  */
 function isOperation(string $value): bool
 {
-    // TODO: Benchmark
-    return \in_array($value, ['query', 'mutation', 'subscription'], true);
+    return $value === 'query' || $value === 'mutation' || $value === 'subscription';
```
fubhy: Good catch.

src/Language/utils.php
crisu83: There is still an issue with this, but I created a separate issue for this (#193).

src/Language/SourceLocation.php
crisu83: @fubhy could you make a separate issue for this?

PR #185, opened 2018-04-06
Parser optimization

crisu83:
Restructured both the AST and node builders to increase their performance by moving away from the builder pattern. This degrades code quality a bit (because the classes are quite big), but I think performance is most important when it comes to parsing. Besides I like the new `NodeBuilder` more than the 30 something builder classes we used to have.


crisu83:
I'm going to merge this because it's green.

PR #188, opened 2018-04-06
Optimize lexer

fubhy:


src/Language/Lexer.php
```diff
@@ -52,20 +46,25 @@ class Lexer implements LexerInterface
...
-    public function __construct(array $readers)
+    public function __construct(TokenReaderInterface $reader)
     {
         $startOfFileToken = new Token(TokenKindEnum::SOF);
 
-        foreach ($readers as $reader) {
-            $reader->setLexer($this);
-        }
+        $this->reader    = $reader;
+        $this->reader->setLexer($this);
```
crisu83: I would prefer swapping the two above lines.

src/Language/LexerInterface.php
```diff
@@ -26,8 +26,9 @@ public function lookahead(): Token;
     public function getBody(): string;
 
     /**
-     * @param string     $name
+     * @param string $name
      * @param mixed|null $default
+     *
```
crisu83: I haven't had an empty line here, but we could consider adding it to the code style included in the project.

src/Language/TokenReaderInterface.php
```diff
@@ -0,0 +1,37 @@
...
+    public function read(
+      int $code,
+      int $pos,
+      int $line,
+      int $col,
+      Token $prev
+    ): Token;
+
+    /**
+     * @param \Digia\GraphQL\Language\LexerInterface $lexer
```
crisu83: I haven't used fully-qualified class names in docblocks, but maybe we should?

crisu83: As long as we're consistent everything goes.

src/Language/TokenReader.php
```diff
@@ -0,0 +1,777 @@
...
+
+    /**
+     * The lexer owning this token reader.
+     *
+     * @var LexerInterface
+     */
+    protected $lexer;
+
+    /**
+     * {@inheritdoc}
```
crisu83: We have used `@inheritdoc`, but I guess this is also as correct. However, we should be consistent about this.

PR #190, opened 2018-04-06
WIP Format code according to PSR-1 & PSR-2.

fubhy:


fubhy:
Let's re-do this from scratch pre-1.0.

PR #191, opened 2018-04-07
WIP: Execution resolve test

hungneox:


tests/Functional/Execution/ResolveTest.php
```diff
@@ -0,0 +1,108 @@
...
+     * @throws \Digia\GraphQL\Error\InvariantException
+     */
+    public function testDefaultFunctionPassesArgsAndContext()
+    {
+        $schema = $this->createTestSchema([
+            'type' => Int(),
+            'args' => ['addend1' => ['type' => Int()]],
+        ]);
+
+        $source = new class(700)
```
hungneox: @crisu83 this is from `graphql-js` test. Do we support field resolver like this?

crisu83: Yes we do.

PR #195, opened 2018-04-08
Experiment: Lex using regex

crisu83:
Refactored the token reader to use regular expressions instead of character codes. This *should* increase the lexing performance. However, I couldn't get the string related regular expressions to work, so it would be nice if @fubhy or @Jalle19 can take a look and see if you can get it to work. Please also note that this feature is far from done because it's missing almost all error reporting.

Based on the `lexer-unit-tests` branch.


crisu83:
Closing this because we're done with the Parser and Lexer for now. :)

PR #196, opened 2018-04-09
Lexer optimizations

crisu83:
Further optimizations to the lexer (and some other things).


crisu83:
@hungneox I actually changed the `Token` back to using protected fields to make it immutable.

crisu83:
I think we're good to merge. With PHP 7.1 we're approx. 1.5x slower than the competition and with PHP 7.2 are very close to the competition. We've also discussed this a lot with the other contributors and we don't think that parsing the most important thing. That said we don't really want to give up too much of the code quality in order to increase the performance. All the changes done in this branch were mainly also done to make the code easier to maintain.

PR #199, opened 2018-04-11
Remove ExecutionStrategy abstract class

hungneox:
- Unify ExecutionStrategy and ExecutorExecutionStrategy into one class.


src/Execution/ExecutionContext.php
```diff
@@ -180,12 +180,12 @@ public function getFragments()
     /**
      * Create proper ExecutionStrategy when needed
      *
-     * @return ExecutionStrategy
+     * @return Executor
      */
-    public function getExecutionStrategy(): ExecutionStrategy
+    public function getExecutor(): Executor
     {
```
crisu83: Call this createExecutor

src/Execution/Executor.php
```diff
@@ -78,10 +78,47 @@ public function __construct(
         $this->rootValue = $rootValue;
     }
 
+
     /**
-     * @return array|null
+     * @return ?array
```
crisu83: Afaik array|null is correct

PR #201, opened 2018-04-12
Remove configuration concept

crisu83:
After I changed the AST nodes to use constructor params when I was optimizing the parser I realized that the same practice could be applied to project-wide. That is why I refactored both the types and schema to use constructor params and removed the config package completely.


src/Type/Definition/FieldsTrait.php
```diff
@@ -77,14 +80,14 @@ protected function buildFieldMap($fieldsOrThunk): array
...
-                \sprintf('%s.%s field config must be an array', $this->getName(), $fieldName)
+                \sprintf('%s.%s field config must be an array', $this->name, $fieldName)
             );
 
             invariant(
                 !isset($fieldConfig['isDeprecated']),
                 \sprintf(
                     '%s.%s should provide "deprecationReason" instead of "isDeprecated".',
-                    $this->getName(),
+                    $this->name,
```
hungneox: @crisu83 Why did you change `$this->getName()` to `$this->name`? :D Plus there is no `$name` attribute in this trait.

src/Type/Definition/FieldsTrait.php
```diff
@@ -94,15 +97,22 @@ protected function buildFieldMap($fieldsOrThunk): array
...
                         toString($fieldConfig['resolve'])
                     )
                 );
             }
 
-            $fieldConfig['name'] = $fieldName;
-            $fieldMap[$fieldName] = new Field($fieldConfig);
+            $fieldMap[$fieldName] = new Field(
+                $fieldName,
+                $fieldConfig['description'] ?? null,
```
hungneox: I think maybe we wnat to get rid of `ConfigAwareTrait` but this kind of workaround smells bad to me :-?

src/Type/Definition/Field.php
```diff
@@ -2,19 +2,51 @@
...
+
+    /**
+     * Field constructor.
+     *
+     * @param string                                                       $name
+     * @param null|string                                                  $description
+     * @param TypeInterface|OutputTypeInterface|WrappingTypeInterface|null $type
+     * @param Argument[]|array[]                                           $arguments
+     * @param callable|null                                                $resolve
+     * @param null|string                                                  $deprecationReason
```
hungneox: I think if you let PHPStorm generate phpdoc it should be `string|null`.

src/Type/Definition/FieldsTrait.php
crisu83: I fixed this.

src/Type/Definition/EnumType.php
```diff
@@ -21,7 +21,7 @@
  *
  * Example:
  *
- *     $RGBType = new GraphQLEnumType([
+ *     $RGBType = new newEnumType([
```
fubhy: That seems wrong.

src/Type/Definition/EnumType.php
```diff
@@ -209,7 +199,7 @@ protected function buildValues(array $valueMap): array
                 isAssocArray($valueConfig),
                 \sprintf(
                     '%s.%s must refer to an object with a "value" key representing an internal value but got: %s',
-                    $this->getName(),
```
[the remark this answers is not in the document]
crisu83: Only for traits.

src/Type/Definition/EnumType.php
```diff
@@ -219,7 +209,7 @@ protected function buildValues(array $valueMap): array
                 !isset($valueConfig['isDeprecated']),
                 \sprintf(
                     '%s.%s should provided "deprecationReason" instead of "isDeprecated".',
-                    $this->getName(),
+                    $this->name,
```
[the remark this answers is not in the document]
crisu83: I'd prefer to use fields directly where possible, in traits we cannot do that for obvious reasons.

PR #203, opened 2018-04-12
Introduce subscribe method to Field

crisu83:
I noticed that we didn't have a subscribe method for fields so I added it based on the reference impl.

### DIFF
--- a/src/Type/Definition/Field.php
+++ b/src/Type/Definition/Field.php
@@ -9,8 +9,6 @@ use Digia\GraphQL\Language\Node\FieldDefinitionNode;
 
 class Field implements ASTNodeAwareInterface, ArgumentsAwareInterface
 {
-    // TODO: Subscription support
-
     use NameTrait;
     use DescriptionTrait;
     use TypeTrait;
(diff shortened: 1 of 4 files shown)


src/Type/Definition/Field.php
```diff
@@ -36,17 +40,46 @@ public function __construct(
         ?string $description,
         ?TypeInterface $type,
         array $arguments,
-        ?callable $resolve,
+        ?callable $resolveCallback,
```
crisu83: Renamed for consistency.

src/Type/Definition/ObjectType.php
```diff
@@ -109,13 +109,6 @@ public function __construct(
         $this->extensionAstNodes = $extensionASTNodes;
 
         invariant(null !== $this->name, 'Must provide name.');
-
-        if (null !== $this->isTypeOfCallback) {
-            invariant(
-                \is_callable($this->isTypeOfCallback),
-                \sprintf('%s must provide "isTypeOf" as a function.', $this->name)
-            );
-        }
```
crisu83: This is now useless because the callback can only be of type `function`.

PR #204, opened 2018-04-13
Missing type hints

hungneox:
- Add missing type-hints and PhpDoc

### DIFF
--- a/src/Execution/Execution.php
+++ b/src/Execution/Execution.php
@@ -29,9 +29,9 @@ class Execution implements ExecutionInterface
     /**
      * @param Schema        $schema
      * @param DocumentNode  $documentNode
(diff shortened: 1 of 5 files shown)


src/Execution/Execution.php
```diff
@@ -29,9 +29,9 @@ public function __construct(ExecutionContextBuilder $contextBuilder)
     /**
      * @param Schema        $schema
      * @param DocumentNode  $documentNode
-     * @param null          $rootValue
+     * @param object|array  $rootValue
```
crisu83: Can this be something else?

hungneox: I'm not really sure, but in all the tests, it can only be array or object.

PR #205, opened 2018-04-13
Add missing functions to type definitions

crisu83:
Add missing type definition functions such as `newField`, `newInputField` and `newArgument`.


src/Schema/Extension/ExtensionContext.php
```diff
@@ -357,9 +355,7 @@ protected function extendFieldMap(TypeInterface $type): array
                 'description'       => $field->getDescription(),
                 'deprecationReason' => $field->getDeprecationReason(),
                 'type'              => $this->extendFieldType($field->getType()),
-                'args'              => keyMap($field->getArguments(), function (Argument $argument) {
-                    return $argument->getName();
-                }),
+                'args'              => $field->getRawArguments(),
```
crisu83: This was actually a bug.

src/Type/Definition/ArgumentsTrait.php
```diff
@@ -23,36 +42,38 @@ public function hasArguments(): bool
...
-                    $argument['defaultValue'] ?? null,
-                    $argument['description'] ?? null,
-                    $argument['astNode'] ?? null
-                );
+        $arguments = [];
+
+        foreach ($rawArguments as $argumentName => $argumentConfig) {
+            $argumentConfig['name'] = $argumentName;
+
+            $arguments[] = newArgument($argumentConfig);
```
crisu83: After cleaning up the logic in `ExtensionContext` I could simplify this to always assume that arguments are given in the "raw" format (associative array).

PR #206, opened 2018-04-13
Refactor type definition functions

crisu83:
Refactor type definition functions to use destructor syntax (thanks @hungneox).

### DIFF
--- a/src/Schema/Schema.php
+++ b/src/Schema/Schema.php
@@ -58,17 +58,17 @@ class Schema implements SchemaInterface
     /**
      * @var TypeInterface|null
      */
-    protected $query;
+    protected $queryType;
 
     /**
      * @var TypeInterface|null
      */
-    protected $mutation;
+    protected $mutationType;
 
     /**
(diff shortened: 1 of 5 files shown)


src/Schema/Schema.php
```diff
@@ -58,17 +58,17 @@ class Schema implements SchemaInterface
...
      * @var TypeInterface|null
      */
-    protected $mutation;
+    protected $mutationType;
 
     /**
      * @var TypeInterface|null
      */
-    protected $subscription;
+    protected $subscriptionType;
```
hungneox: @crisu83 One thing I want to ask is can we return `ObjectType` instead of `TypeInterface`. Since they cannot be any thing else. And this is only for internal api.

crisu83: We cannot, I cannot recall why, but there was a reason for why it has to be `TypeInterface|null`.

hungneox: I think it's probably because of using container?

PR #209, opened 2018-04-15
Definition printer

crisu83:


src/GraphQL.php
```diff
@@ -254,7 +254,7 @@ public static function execute(
      */
     public static function print(NodeInterface $node): string
     {
-        return static::make(PrinterInterface::class)->print($node);
+        return static::make(NodePrinterInterface::class)->print($node);
```
crisu83: Renamed this for consistency.

src/Language/NodePrinter.php
```diff
@@ -366,7 +368,7 @@ protected function one(?NodeInterface $node): string
      * @param array $nodes
      * @return array
      */
-    protected function many(array $nodes): array
+    protected function printMany(array $nodes): array
```
crisu83: Renamed these methods based @hungneox 's suggestion.

src/Type/Definition/DefinitionInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Digia\GraphQL\Type\Definition;
+
+/**
+ * Tagging interface for GraphQL definitions (schema, types, directive, etc.).
+ */
+interface DefinitionInterface
+{
+}
```
crisu83: Introduce this tagging interface to be able to type-hint classes within the `DefinitionPrinter`

src/Type/Definition/TypeInterface.php
```diff
@@ -2,15 +2,10 @@
...
 {
     /**
      * @return string|null
      */
     public function getName(): ?string;
-
-    /**
-     * @return string|null
-     */
-    public function getDescription(): ?string;
```
crisu83: This was actually a bug, `ListType` and `NonNullType` cannot have a description.

src/Type/Definition/DescriptionAwareInterface.php
```diff
@@ -0,0 +1,11 @@
...
+
+namespace Digia\GraphQL\Type\Definition;
+
+interface DescriptionAwareInterface
+{
+    /**
+     * @return null|string
+     */
+    public function getDescription(): ?string;
+}
```
crisu83: Introduce this interface to be able to check whether or not the class has a description.

src/Schema/SchemaInterface.php
```diff
@@ -9,6 +9,8 @@
 use Digia\GraphQL\Type\Definition\ObjectType;
 use Digia\GraphQL\Type\Definition\TypeInterface;
 
+// TODO: Consider removing this interface.
+
 interface SchemaInterface
```
crisu83: I think we should remove this interface. I don't see why you'd want to replace it. Besides, it's not registered with the container anyway. However, I think I'll do this in a separate PR.

src/Language/NodePrinter.php
```diff
@@ -30,11 +32,11 @@
 use Digia\GraphQL\Language\Node\VariableNode;
 use function Digia\GraphQL\Util\toString;
 
-class Printer implements PrinterInterface
+class NodePrinter implements NodePrinterInterface
```
crisu83: After renaming this class I realized that we should probably move both this and the `NodeBuilder` into the `Node` namespace as it uses quite a few classes from that namespace and the class name also starts with the word `Node`.

fubhy: Agreed.

src/Schema/SchemaInterface.php
fubhy: If it's not swappable anyways there is no point in having it indeed.

crisu83:
I would actually like to merge this ASAP because it contains a lot of all-around improvements. I created a separate issue for the test cases (#211). @fubhy @hungneox are you ok with this?

Issue #212, opened 2018-04-24
Remove SchemaInterface


crisu83:
The `SchemaInterface` is redundant as the schema cannot be replaced in the DI container (this is by design).

crisu83:
This issue was fixed in #213.

PR #213, opened 2018-04-24
Remove SchemaInterface

crisu83:
See #212.


crisu83:
@Jalle19 Yeah I noticed that too. It's fixed now.

PR #216, opened 2018-05-06
Remove Cache package

crisu83:
Remove Cache package because it's not used anymore.


crisu83:
No need to review this, as I only removed some unused classes.

PR #217, opened 2018-05-06
Refactor execution package

crisu83:
Reviewed the Execution package as a whole and removed at least most of the IDE warnings. 

Feel free to question any of my changes.


src/Execution/Executor.php
```diff
@@ -588,27 +536,27 @@ private function completeValue(
...
         ResolveInfo $info,
-        $path,
+        array $path,
         &$result
     ) {
         $runtimeType = $returnType->resolveType($result, $this->context->getContextValue(), $info);
 
         if (null === $runtimeType) {
-            //@TODO Show warning
+            // TODO: Display warning
```
hungneox: Why you change `@TODO` to `TODO:`?

src/Execution/Executor.php
```diff
@@ -653,67 +589,81 @@ private function completeAbstractValue(
...
 
         if (!$runtimeType instanceof ObjectType) {
             $parentTypeName = $info->getParentType()->getName();
             $fieldName      = $info->getFieldName();
 
             throw new ExecutionException(
-                "Abstract type {$returnTypeName} must resolve to an Object type at runtime " .
-                "for field {$parentTypeName}.{$fieldName} with " .
-                'value "' . $result . '", received "{$runtimeTypeName}".'
+                \sprintf(
```
hungneox: I intentionally avoid using `sprintf` for long message because it's very hard to know which variable in which placeholder.

src/Execution/Execution.php
```diff
@@ -65,8 +50,102 @@ public function execute(
...
+        $errors = $context->getErrors();
+
+        return new ExecutionResult($data, $errors);
+    }
+
+    /**
+     * @param Schema        $schema
+     * @param DocumentNode  $documentNode
+     * @param               $rootValue
+     * @param               $contextValue
```
hungneox: Probably missing type-hint for $rootValue and $contextValue (mixed).

src/Execution/Executor.php
```diff
@@ -507,21 +339,137 @@ public function completeValueWithLocatedError(
...
+            }
+
+            $doesContainPromise = $doesContainPromise || $this->isPromise($result);
+
+            $finalResults[$fieldName] = $result;
+        }
+
+        if ($doesContainPromise) {
+            $keys    = array_keys($finalResults);
+            $promise = \React\Promise\all(array_values($finalResults));
```
hungneox: @crisu83 If you want to remove all \React\Promise namespace, we can probably make an alias for this.

`using \React\Promise\all as promiseAll`

src/Execution/ExecutionContextBuilder.php
```diff
@@ -1,100 +0,0 @@
...
-use Digia\GraphQL\Language\Node\DocumentNode;
-use Digia\GraphQL\Language\Node\NodeKindEnum;
-use Digia\GraphQL\Language\Node\OperationDefinitionNode;
-use Digia\GraphQL\Schema\Schema;
-
-/**
- * Class ExecutionContextBuilder
- * @package Digia\GraphQL\Execution
- */
-class ExecutionContextBuilder
```
hungneox: @crisu83 I think to have this Builder is somehow easier to write unit test.

src/Execution/Execution.php
```diff
@@ -65,8 +50,102 @@ public function execute(
...
+     * @param               $rawVariableValues
```
Jalle19: Same here

src/Execution/FieldCollector.php
```diff
@@ -106,6 +112,7 @@ protected function shouldIncludeNode(NodeInterface $node): bool
 
         $include = coerceDirectiveValues(IncludeDirective(), $node, $contextVariables);
 
+        /** @noinspection IfReturnReturnSimplificationInspection */
```
Jalle19: I would change the return to a ternary instead

src/Execution/ExecutionContextBuilder.php
crisu83: There was no tests and removing the builder is consistent with rest of the project.

src/Execution/Executor.php
crisu83: Again, consistency and also readaility.

src/Execution/Executor.php
crisu83: Consistency.

src/Execution/Executor.php
hungneox: I don't think it's easier to read :D If consistency open a door for potential issue I would I think again. Context is of usage is also important.

crisu83: A potential issue with an error message? I wouldn't sacrifice consistency for that.

src/Execution/Executor.php
crisu83: `@TODO` is usually used inside docblocks.

src/Execution/Executor.php
```diff
@@ -61,60 +61,51 @@ class Executor
...
         $objectType = $this->getOperationType($schema, $operation);
 
         $fields               = [];
         $visitedFragmentNames = [];
+
         try {
             $fields = $this->fieldCollector->collectFields(
                 $objectType,
-                $this->operation->getSelectionSet(),
+                $this->context->getOperation()->getSelectionSet(),
```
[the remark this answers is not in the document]
crisu83: Fixed.

src/Execution/Execution.php
crisu83: Fixed.

src/Execution/Executor.php
crisu83: Fixed.

src/Execution/ExecutionContextBuilder.php
crisu83: I would still remove the builder because I also removed all the other context builders to reduce complexity.

src/Execution/Execution.php
crisu83: Fixed.

src/Execution/FieldCollector.php
crisu83: I didn't want to do that because new execution paths could be introduced in the future.

src/Execution/Executor.php
crisu83: We have numerous similar `sprintf` statements in the `Schema` package so changing this back won't really solve the issue alone. I would change this for consistency now and open a proposal for not using `sprintf` for very long error messages.

src/Execution/ExecutionContextBuilder.php
crisu83: Also, I don't really see the need for unit testing the builder. We have quite a lot of "black-box" tests for this already.

Issue #218, opened 2018-05-06
Fix possible bugs noticed by Scrutinizer


crisu83:
Currently, Scrutinizer finds 194 issues with the project:
https://scrutinizer-ci.com/g/digiaonline/graphql-php/issues/master

We should go through all of these and fix any potential bugs.

Jalle19:
I'll handle some of these

crisu83:
@Jalle19 approved your PR.

PR #221, opened 2018-05-08
Fix various code issues

Jalle19:
Related to #218

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,6 @@
             "./tests/Functional/Execution/testClasses.php",
             "./tests/Functional/Validation/errors.php",
             "./tests/Functional/Validation/harness.php",
-            "./tests/Functional/QueryTest.php",
             "./tests/Functional/starWarsData.php",
             "./tests/Functional/starWarsSchema.php",
             "./tests/utils.php"
--- a/src/Util/ValueHelper.php
+++ b/src/Util/ValueHelper.php
@@ -50,8 +50,8 @@ class ValueHelper
     }
 
     /**
-     * @param $valueA
-     * @param $valueB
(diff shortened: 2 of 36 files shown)


composer.json
```diff
@@ -60,7 +60,6 @@
             "./tests/Functional/Execution/testClasses.php",
             "./tests/Functional/Validation/errors.php",
             "./tests/Functional/Validation/harness.php",
-            "./tests/Functional/QueryTest.php",
```
crisu83: Good catch.

src/Util/ValueHelper.php
```diff
@@ -50,8 +50,8 @@ function compareArguments(array $argumentsA, array $argumentsB): bool
     }
 
     /**
-     * @param $valueA
-     * @param $valueB
+     * @param ValueNodeInterface $valueA
+     * @param ValueNodeInterface $valueB
      * @return bool
      */
     public function compareValues($valueA, $valueB): bool
```
crisu83: Could these even be type-hinted? @Jalle19

src/Validation/Conflict/Map.php
```diff
@@ -28,8 +28,8 @@ public function __construct(array $values = [])
 
 
     /**
-     * @param $key
-     * @param $value
+     * @param string $key
+     * @param mixed  $value
      */
     public function set($key, $value): void
```
crisu83: The `$key` could probably be type-hinted. @Jalle19

src/Util/ValueHelper.php
Jalle19: Yes, I had to rebase on master and this change got lost

crisu83:
Rebased on `master`.

PR #226, opened 2018-05-15
Fix analysis issues

crisu83:
Fix all of the remaining level 2 PHPStan analysis errors in the `src` directory.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,11 +48,11 @@
     "scripts": {
         "test": [
             "phpunit",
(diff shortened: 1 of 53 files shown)


Jalle19:
Can you bump the analysis level in composer.json too?

crisu83:
@Jalle19 not until the tests are fixed too. :(

crisu83:
There was quite a few issues, we could indeed consider excluding tests from the static analysis.

PR #232, opened 2018-06-06
Ensure return types of parse() and GraphQL::parse() match

spawnia:
Small and straightforward fix. GraphQL::parse() returns DocumentNode so parse() should do so as well.


crisu83:
Looks good to me.

PR #237, opened 2018-06-17
Parse partial

### LINKED ISSUE #234: [Feat] Support parsing partial SDL
We have decided to add support for parsing a partial SDL. An example of this is parsing a field definitions from `foo: String!`. 

The initial plan for implement this is the following:

- [ ] Rename all internal methods on `Parser` to e.g. `lexValueLiteral` (instead of `parse...`)
- [ ] Introduce public parse methods that all have the following signature `parse...($source, array $options = [])` (not that these methods do not need to be added to the public API)

crisu83:
Add initial support for partial parsing (fixes #234).

### DIFF
--- a/src/Language/Parser.php
+++ b/src/Language/Parser.php
@@ -116,54 +116,6 @@ class Parser implements ParserInterface
     }
 
     /**
-     * Given a string containing a GraphQL value (ex. `[42]`), parse the AST for
-     * that value.
-     * Throws GraphQLError if a syntax error is encountered.
-     *
-     * This is useful within tools that operate upon GraphQL Values directly and
-     * in isolation of complete GraphQL documents.
-     *
-     * @inheritdoc
-     * @throws SyntaxErrorException
-     * @throws InvariantException
-     * @todo Deprecate this method in favor of `parsePartial`
-     */
-    public function parseValue($source, array $options = []): NodeInterface
-    {
-        $this->lexer = $this->createLexer($source, $options);
-
-        $this->expect(TokenKindEnum::SOF);
-        $value = $this->lexValueLiteral();
-        $this->expect(TokenKindEnum::EOF);
-
-        return $value;
-    }
-
-    /**
-     * Given a string containing a GraphQL Type (ex. `[Int!]`), parse the AST for
-     * that type.
-     * Throws GraphQLError if a syntax error is encountered.
-     *
-     * This is useful within tools that operate upon GraphQL Types directly and
-     * in isolation of complete GraphQL documents.
-     *
-     * @inheritdoc
-     * @throws SyntaxErrorException
-     * @throws InvariantException
-     * @todo Deprecate this method in favor of `parsePartial`
-     */
-    public function parseType($source, array $options = []): TypeNodeInterface
-    {
-        $this->lexer = $this->createLexer($source, $options);
-
-        $this->expect(TokenKindEnum::SOF);
-        $type = $this->lexTypeReference();
-        $this->expect(TokenKindEnum::EOF);
-
-        return $type;
-    }
-
-    /**
      * Converts a name lex token into a name parse node.
      *
      * @return NameNode
@@ -362,14 +314,14 @@ class Parser implements ParserInterface
          */
         $parseType = function (): TypeNodeInterface {
             $this->expect(TokenKindEnum::COLON);
-            return $this->lexTypeReference();
+            return $this->lexType();
         };
 
         return new VariableDefinitionNode(
             $this->lexVariable(),
             $parseType(),
             $this->skip(TokenKindEnum::EQUALS)
-                ? $this->lexValueLiteral(true)
+                ? $this->lexValue(true)
                 : null,
             $this->createLocation($start)
         );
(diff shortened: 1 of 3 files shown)


src/Language/Parser.php
```diff
@@ -61,6 +61,25 @@ class Parser implements ParserInterface
      */
     protected $lexer;
 
+    /**
```
spawnia: I would suggest adding a comment that outlines the purpose of this magic function.

src/Language/Parser.php
```diff
@@ -62,65 +62,57 @@ class Parser implements ParserInterface
...
+     * @return mixed
      *
-     * @inheritdoc
-     * @throws SyntaxErrorException
-     * @throws \ReflectionException
      * @throws InvariantException
+     * @throws SyntaxErrorException
      */
-    public function parse($source, array $options = []): DocumentNode
+    public function __call(string $name, array $arguments)
```
spawnia: Actually, now that i thought about it some more, i do think that using a magic method here is not the best way to go. It does solve the issue with very few lines of code, but on the other hand makes the code less clean.

crisu83: I would normally agree, however in this case we would need 30+ parse methods, so I opted for this approach instead. This is also not really part of the public api so I think it’s acceptable.

crisu83: @spawnia thoughts? Can we go with this solution now, as I’d like to keep it simple. We can change it later if need be.

src/Language/Parser.php
```diff
@@ -62,65 +62,57 @@ class Parser implements ParserInterface
...
+     * @throws SyntaxErrorException
      */
-    public function parseType($source, array $options = []): TypeNodeInterface
+    protected function parsePartial(callable $lexCallback, $source, array $options = [])
     {
         $this->lexer = $this->createLexer($source, $options);
 
         $this->expect(TokenKindEnum::SOF);
-        $type = $this->parseTypeReference();
+        $node = \call_user_func($lexCallback, $source, $options);
```
Jalle19: You can do `$lexCallback($source, $options)`, no need for call_user_func

Jalle19:
@crisu83 @spawnia we could add `@method` phpdoc to the class to simulate having individual methods

spawnia:
@Jalle19 nice idea, sounds like it will be the best of both worlds!

src/Language/Parser.php
crisu83: We should probably change the use of callbacks everywhere.

crisu83:
I tried the `@method` approach, but there is no way to have it behave correctly in our case. If we put the `@method` mentions on `ParserInterface` it will require `Parser` to implement such methods, which it doesn't, because it uses `__call`. If we, on the other hand, put the `@method` mentions on the `Parser` it will work, but it feels weird that we "make" the parser instance from its interface, yet we would need to implicitly name the implementation if we'd want the auto-complete to work.

The only option to the current approach is to add all the methods to `ParserInterface` and implement them without using `__call`. The only problem with this is that there are 32 methods that need to be implemented, and I might have even missed a few. 🤔

Jalle19:
Are you sure they have to be implemented? Maybe it's just a PhpStorm thing?

olivernybroe:
@crisu83 that is a phpstorm thing. Php does not use the comments.
Phpstorm wants you to add all the methods to both classes, but you can just ignore that and only add it to the interface. (maybe just disable that inspection on the implementation?)

You can also solve the autocompletion with a phpstorm meta file.

crisu83:
@olivernybroe @Jalle19 @spawnia Alright, I pushed the `@method` mentions for `ParserInterface` and ignored the IDE error in the implementation. Please check and let's get this merged.

src/Language/ParserInterface.php
```diff
@@ -3,12 +3,46 @@
...
+ * @method parseEnumValueDefinition($source, array $options = []): EnumValueDefinitionNode
+ * @method parseInputObjectTypeDefinition($source, array $options = []): InputObjectTypeDefinitionNode
+ * @method parseScalarTypeExtension($source, array $options = []): ScalarTypeExtensionNode
+ * @method parseObjectTypeExtension($source, array $options = []): ObjectTypeExtensionNode
+ * @method parseInterfaceTypeExtension($source, array $options = []): InterfaceTypeExtensionNode
+ * @method parseUnionTypeExtension($source, array $options = []): UnionTypeExtensionNode
+ * @method parseEnumTypeExtension($source, array $options = []): EnumTypeExtensionNode
+ * @method parseInputObjectTypeExtension($source, array $options = []): InputObjectTypeExtensionNode
+ * @method parseDirectiveDefinition($source, array $options = []): DirectiveDefinitionNode
+ */
```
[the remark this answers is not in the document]
spawnia: I am not exactly sure - looking through Language\Node i counted smth like 55 classes that end in Node, the methods account to 32.

tests/Functional/Language/ParserTest.php
```diff
@@ -20,6 +29,56 @@
...
+
+        /** @var DirectiveNode $directiveNode */
+        $directiveNode = $parser->parseDirective('@foo(bar: String, baz: Int)');
+        $directiveArgs = $directiveNode->getArguments();
+
+        $this->assertInstanceOf(DirectiveNode::class, $directiveNode);
+        $this->assertEquals('foo', $directiveNode->getNameValue());
+        $this->assertEquals('bar', $directiveArgs[0]->getNameValue());
+        $this->assertEquals('baz', $directiveArgs[1]->getNameValue());
+    }
```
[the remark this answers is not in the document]
spawnia: I guess since we are just delegating through a magic method, a few test cases should suffice.

PR #244, opened 2018-06-18
Link to read-only README of graphql-relay-php

hugovk:
Not the edit page.


crisu83:
Indeed, thanks @hugovk!

PR #250, opened 2018-07-06
 Replace symfony/polyfill-mbstring with a hard ext-mbstring requirement

### LINKED ISSUE #247: Make the mbstring extension a hard requirement
I made a simple API using this library and deployed it to both Heroku and a server on UpCloud. In both cases, performance was abysmally bad - on Heroku each GraphQL request consistently took over 10 seconds while on UpCloud they were closer to 5 seconds.

After install the mbstring extension, the response time on UpCloud dropped to 25 ms.

Jalle19:
Fixes #247


Jalle19:
I tried improving performance with the polyfill yesterday but it seemed pretty futile. The polyfill is simply too slow for our needs.

PR #252, opened 2018-07-06
 Bump the phpstan analysis level to 4 

Jalle19:
@crisu83 this actually fixes some bugs so in my opinion it's important that we get it in.

### DIFF
--- a/src/Type/Coercer/IntCoercer.php
+++ b/src/Type/Coercer/IntCoercer.php
@@ -8,6 +8,8 @@ class IntCoercer extends AbstractCoercer
 {
     /**
      * @inheritdoc
+     *
+     * @throws InvalidTypeException
      */
     public function coerce($value)
     {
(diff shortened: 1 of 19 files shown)


src/Type/Coercer/IntCoercer.php
```diff
@@ -19,13 +21,13 @@ public function coerce($value)
...
 
-        if (!\is_int($value) || $value > PHP_INT_MAX || $value < PHP_INT_MIN) {
+        if (!\is_numeric($value) || $value > PHP_INT_MAX || $value < PHP_INT_MIN) {
             throw new InvalidTypeException(
                 \sprintf('Int cannot represent non 32-bit signed integer value: %s', $value)
             );
         }
 
-        $intValue   = (int)$value;
+        $intValue   = $value;
```
spawnia: That assignment is no longer needed, right?

Jalle19: You're probably right

tests/Unit/Util/ValueConverterTest.php
```diff
@@ -0,0 +1,43 @@
...
+     * @throws \Digia\GraphQL\Error\ConversionException
+     * @throws \Digia\GraphQL\Error\InvariantException
+     * @throws \Digia\GraphQL\Error\SyntaxErrorException
+     */
+    public function testIdType(): void
+    {
+        $idType = ID();
+
+        // Ensure numerical IDs become Ints
+        $this->assertInstanceOf(IntValueNode::class, $this->valueConverter->convert('45', $idType));
```
crisu83: I would just call `ID()` multiple times for less code.

Jalle19: Fixed

Jalle19:
Scrutinizer complains about ValueConverter, `The condition $type === ID() is always false. `.

Jalle19:
Maybe we should remove the special case handling for the ID type and just let it be a string?

Jalle19:
I made a separate issue for the ID type issues so we can get this merged: https://github.com/digiaonline/graphql-php/issues/255

Issue #239, opened 2018-06-17
Fix ambiguity with null variable values and default values


crisu83:
We need to ensure that null variable values and default values are handled correctly. The reference implementation can be found here: https://github.com/graphql/graphql-js/pull/1274/files

crisu83:
@Jalle19 do you know if this issue is done?

Jalle19:
I don't know, but I think the issue still exists

crisu83:
@Jalle19 I think this works as it should in our implementation. I vote for closing this issue and re-opening it if it resurfaces. What do you think?

Jalle19:
Sure

PR #258, opened 2018-07-12
Add initial tests for DefinitionPrinter (including bug fixes)

### LINKED ISSUE #211: DefinitionPrinter tests
We need to port the test cases for the `DefinitionPrinter` class. The reference implementation can be found here: https://github.com/graphql/graphql-js/blob/master/src/utilities/__tests__/schemaPrinter-test.js

Jalle19:
Fixes #211 

Some parts are still missing:

- [ ] string field test with an escaped default value
- [ ] string field with a default null value (requires #239)
- [ ] string field with non-null integer argument
- [ ] support for printing directives
- [ ] support for printing the introspection schema

### DIFF
--- a/tests/Functional/Schema/DefinitionPrinterTest.php
+++ b/tests/Functional/Schema/DefinitionPrinterTest.php
@@ -1,0 +1,566 @@
+<?php
+
+namespace Digia\GraphQL\Test\Functional\Schema;
+
+use Digia\GraphQL\Schema\DefinitionPrinter;
+use Digia\GraphQL\Test\TestCase;
+use function Digia\GraphQL\buildSchema;
+use function Digia\GraphQL\Language\dedent;
+
+/**
+ * Class DefinitionPrinterTest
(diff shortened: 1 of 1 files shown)


tests/Functional/Schema/DefinitionPrinterTest.php
```diff
@@ -0,0 +1,566 @@
...
+     * @throws \Digia\GraphQL\Error\InvariantException
+     * @throws \Digia\GraphQL\Error\PrintException
+     */
+    private function assertPrintedSchemaEqualsBuiltSchema(string $source): void
+    {
+        $schema = buildSchema($source);
+
+        $printer = new DefinitionPrinter();
+
+        $this->assertEquals($source, $printer->printSchema($schema));
```
spawnia: Would change to `assertSame` since that uses ===

Jalle19:
@spawnia I changed it to use `assertSame`

crisu83:
@Jalle19 Can you please create another issue for the remaining issues? Thanks!

PR #259, opened 2018-07-12
Add unit tests and improved phpdoc for ValueHelper, refactor some helpers to be static

Jalle19:


src/Util/ValueHelper.php
```diff
@@ -9,8 +9,8 @@
 class ValueHelper
 {
     /**
-     * @param array $argumentsA
-     * @param array $argumentsB
+     * @param ArgumentNode[] $argumentsA
+     * @param ArgumentNode[] $argumentsB
      * @return bool
      */
     public function compareArguments(array $argumentsA, array $argumentsB): bool
```
spawnia: Can we make those static?

Jalle19: The class is made via the container IIRC, but I guess we can change it

Jalle19:
@spawnia @crisu83 I refactored some of the helpers to have static methods only. This allows us to remove `UtilityProvider` which IMO is a bit of an over-abstraction (it seems unlikely in my opinion that someone would need to swap helpers for their own implementations). This should also provide some marginal performance improvements since all related helper calls are now standard static method calls - previously each invocation meant the helper class had to be resolved through the container.

hungneox:
@Jalle19 Just want to ask about the static method. Is there any advantage of using static method over namespace function? IMO, if we don't have to share any attributes between methods then namespace function is fine. And as we're using PHP7, namespace functions do not pollute global namespace either.

Jalle19:
@hungneox we could add functions, but I don't think it helps much. Furthermore functions can't be autoloaded so the more function files we can get rid of from composer.json the better.

crisu83:
I'm open to considering merging this change. I might point out that the reason why we had the `UtilityProvider` was to allow the various utility implementations to be overridden through the DI container. That said, I don't really think it's necessary to allow these to be overridden. Thoughts?

crisu83:
@Jalle19 Fair enough. If everything works we can merge this.

PR #260, opened 2018-07-12
Change a bunch of assertions from assertEquals to assertSame

spawnia:
Luckily, nothing did break because of this, so the code seems fine as is!
Still, going forward we should aim to use assertSame wherever possible,
as it uses triple equal sign comparison.

### DIFF
--- a/tests/Functional/Execution/ExecutionTest.php
+++ b/tests/Functional/Execution/ExecutionTest.php
@@ -440,7 +440,7 @@ SRC;
                 ],
                 'c'    => 'Cherry'
             ]
-        ], $executionResult)
+        ], $executionResult);
     }
 
     /**
--- a/tests/Functional/Execution/ExecutionTest.php
+++ b/tests/Functional/Execution/ExecutionTest.php
@@ -430,17 +430,17 @@ SRC;
             'data' => [
                 'a'    => 'Apple',
                 'b'    => 'Banana',
-                'c'    => 'Cherry',
                 'deep' => [
                     'b'      => 'Banana',
-                    'c'      => 'Cherry',
                     'deeper' => [
                         'b' => 'Banana',
                         'c' => 'Cherry'
-                    ]
-                ]
+                    ],
+                    'c'      => 'Cherry',
+                ],
+                'c'    => 'Cherry'
             ]
-        ], $executionResult);
+        ], $executionResult)
     }
 
     /**
--- a/tests/Functional/Execution/ExecutionTest.php
+++ b/tests/Functional/Execution/ExecutionTest.php
@@ -77,7 +77,7 @@ class ExecutionTest extends TestCase
         /** @var ExecutionResult $executionResult */
         $result = graphql($schema, $source, $rootValue);
 
-        $this->assertEquals(['data' => ['a' => $rootValue]], $result);
+        $this->assertSame(['data' => ['a' => $rootValue]], $result);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


crisu83:
@spawnia Something seems to be broken, we need to fix the issue before we can merge this.

spawnia:
About `ExecutionTest::testMergeParallelFragments`: If i understand the spec correctly, the actual result is correct! http://facebook.github.io/graphql/June2018/#example-7924b

That means, we should adapt the test case to match the expected order. While the test-case as is defines a nice, logical structure, it does not reflect the order in which the fragments are added in to the query.

spawnia:
And concerning the coercers, any reason why they, of all things, do not have type hints?

tests/Functional/Execution/ExecutionTest.php
```diff
@@ -426,7 +426,7 @@ public function testMergeParallelFragments()
         /** @var ExecutionResult $executionResult */
         $executionResult = graphql($schema, $source);
 
-        $this->assertEquals([
+        $this->assertSame([
```
hungneox: @spawnia You have to modifier the array here in order to make two arrays identical

```
        $this->assertSame([
            'data' => [
                'a'    => 'Apple',
                'b'    => 'Banana',
                'deep' => [
                    'b'      => 'Banana',
                    'deeper' => [
                        'b' => 'Banana',
                        'c' => 'Cherry'
                    ],
                    'c'      => 'Cherry',
                ],
                'c'    => 'Cherry'
            ]
        ], $executionResult);
```

spawnia: Done

spawnia:
@hungneox how about Digia\GraphQL\Test\Unit\Type\Coercer\IntCoercerTest::testSuccessfulCoercion

Seems like the implementation is bugged there?

tests/Functional/Execution/ExecutionTest.php
```diff
@@ -426,21 +426,21 @@ public function testMergeParallelFragments()
...
                         'c' => 'Cherry'
-                    ]
-                ]
+                    ],
+                    'c'      => 'Cherry',
+                ],
+                'c'    => 'Cherry'
             ]
-        ], $executionResult);
+        ], $executionResult)
```
hungneox: @spawnia You missed on semi-colon here

hungneox:
@spawnia What's wrong with this how about Digia\GraphQL\Test\Unit\Type\Coercer\IntCoercerTest::testSuccessfulCoercion 🤔

spawnia:
@hungneox As the test says, the int corcer should return an int, not 2.0

I really do not get why we do not use type hints for the coercer, a place where correctness of types is absolutely crucial.

hungneox:
@spawnia You're probably right. Could you please fix the type-hint in `Type/Coercer/IntCoercer.php` to `public function coerce($value): int`. I will probably fix the rest.

spawnia:
You can edit this PR if you'd like, or maybe make a second PR for adding type hints to the coercers?

hungneox:
@spawnia Sounds good. But you can also get this PR merged first :-?

spawnia:
How about we merge this PR as is? It did its job by uncovering an Issue within the Int Coercer.

It's not like i introduced a bug, i merely corrected the test to actually find the bug. Fixing it is another issue.

hungneox:
@spawnia Sorry, But you have to made all the checks pass first.

spawnia:
I do not agree with that. Better to have a failing test then to have a bug.

hungneox:
@spawnia You could fix the specific case for IntCoercer or change testSuccessfulCoercion back to `assertEquals`. The point is the pipeline has to be green in order to merge, which mean all know checks are passed.

Afterward, you can create another PR for fixing all the Coercer classes.

hugovk:
It's great if you found a bad test which identifies a bug in the code! 

We don't want to merge PRs that end up failing the builds, that means other PRs will fail and make it hard to see the cause.

How about either:

* fixing the test and the code in this PR, or
* fixing the test the code in another PR?

That way everything stays green!

hungneox:
@spawnia It's time to make another PR for the type hint :)

PR #264, opened 2018-07-23
#263 add type hints for coercers

### LINKED ISSUE #263: Fix type-hints for coercer classes
See comments in [#262 comment](https://github.com/digiaonline/graphql-php/pull/260#issuecomment-406031525)

hungneox:
Fixes #263

### DIFF
--- a/src/Type/Coercer/BooleanCoercer.php
+++ b/src/Type/Coercer/BooleanCoercer.php
@@ -9,7 +9,7 @@ class BooleanCoercer extends AbstractCoercer
     /**
      * @inheritdoc
      */
-    public function coerce($value)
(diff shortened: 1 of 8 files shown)


crisu83:
@hungneox Build is failing though.

hungneox:
@crisu83 I'm fixing this. Type hinting automatically casts return values to int or float, but phpstan still complains about it.
```
 Method Digia\GraphQL\Type\Coercer\IntCoercer::coerce() should return  
         int but returns float|int|string.    
```

spawnia:
Hm, now that i think about it some more.. isn't the job of the coercer to come try and come up with a value, no matter what?

PR #266, opened 2018-08-19
Add some intermediate variables so type-hint information isn't lost

Jalle19:
The container's make() method obviously doesn't know what it returns so we have to add @var annotations here and there


Jalle19:
The so called "major" issues Scrutinizer complains about is actually a bug in Scrutinizer: https://github.com/scrutinizer-ci/scrutinizer/issues/274

PR #267, opened 2018-08-19
Add a SourceBuilderInterface along with two basic implementations

Jalle19:
The FileSourceBuilder just does \file_get_contents() on a file, so it's not that useful unless we change the API to accept only Source and not string.

The DirectorySourceBuilder is more interesting since it allows people to split their giant SDL files into multiple logic units (e.g. one for Relay, one for common sorting/ordering stuff, etc.).

### DIFF
--- a/src/Language/FileSourceBuilder.php
+++ b/src/Language/FileSourceBuilder.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Digia\GraphQL\Language;
+
+use Digia\GraphQL\Error\FileNotFoundException;
+use Digia\GraphQL\Error\InvariantException;
+
+/**
+ * Class FileSourceBuilder
+ * @package Digia\GraphQL\Language
+ */
+class FileSourceBuilder implements SourceBuilderInterface
+{
+    /**
+     * @var string
+     */
+    private $filePath;
+
+    /**
+     * FileSourceBuilder constructor.
+     * @param string $filePath
+     */
+    public function __construct(string $filePath)
+    {
+        $this->filePath = $filePath;
+    }
+
+    /**
+     * @inheritdoc
+     *
+     * @throws FileNotFoundException
+     * @throws InvariantException
+     */
+    public function build(): Source
+    {
+        if (!\file_exists($this->filePath) || !\is_readable($this->filePath)) {
+            throw new FileNotFoundException(sprintf('The file %s cannot be found or is not readable', $this->filePath));
+        }
+
+        return new Source(\file_get_contents($this->filePath));
+    }
+}
--- a/src/Language/MultiFileSourceBuilder.php
+++ b/src/Language/MultiFileSourceBuilder.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace Digia\GraphQL\Language;
+
+use Digia\GraphQL\Error\FileNotFoundException;
+use Digia\GraphQL\Error\InvariantException;
+
+/**
+ * Class MultiFileSourceBuilder
+ * @package Digia\GraphQL\Language
+ */
+class MultiFileSourceBuilder implements SourceBuilderInterface
+{
+
+    /**
+     * @var string[]
+     */
+    private $filePaths;
+
+    /**
+     * MultiFileSourceBuilder constructor.
+     * @param string[] $filePaths
+     */
+    public function __construct(array $filePaths)
+    {
+        $this->filePaths = $filePaths;
+    }
+
+    /**
+     * @inheritdoc
+     *
+     * @throws FileNotFoundException
+     * @throws InvariantException
+     */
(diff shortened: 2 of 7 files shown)


Jalle19:
It's based on the minor-fixes branch so it'll need a rebase once that one's gone in. @crisu83 what do you think about this? I've tested it locally in a small project and the directory source builder works as intended.

Jalle19:
Rebased

Jalle19:
The DirectorySourceBuilder could perhaps be more useful if instead of reading a directory it just took an array of absolute paths to files. This way it would be possible to include partial schema files from dependencies (e.g. our Relay implementation) without having to copy-paste definitions.

crisu83:
This is actually more a reader than a builder, am I correct? So maybe we should call it SourceReader instead. What do you think @Jalle19?

Jalle19:
I'll think of something

Jalle19:
@crisu83 I changed `DirectorySourceBuilder` to `MultiFileSourceBuilder`. Should I update the public API too? We need to decide that before I do any README changes.

crisu83:
Let's change the public API later. However, we should use the builder instead of `file_get_contents` in our examples.

Jalle19:
Okay, I'll update the PR

src/Language/SourceBuilderInterface.php
```diff
@@ -0,0 +1,16 @@
...
+ * Interface SourceBuilderInterface
+ * @package Digia\GraphQL\Language
+ */
+interface SourceBuilderInterface
+{
+
+    /**
+     * @return Source
+     */
+    public function buildSource(): Source;
```
crisu83: Call this method `build` for consistency.

tests/Functional/Language/FileSourceBuilderTest.php
```diff
@@ -0,0 +1,34 @@
...
+ */
+class FileSourceBuilderTest extends TestCase
+{
+
+    /**
+     * @expectedException \Digia\GraphQL\Error\InvariantException
+     */
+    public function testFileNotFound(): void
+    {
+        $builder = new FileSourceBuilder('/not/existing/file.graphqls');
```
crisu83: I think it would be better if the builder would be stateless. I would consider changing the build method to accept the path(s) instead.

src/Language/MultiFileSourceBuilder.php
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * @var string[]
+     */
+    private $filePaths;
+
+    /**
+     * MultiFileSourceBuilder constructor.
+     * @param string[] $directoryPaths
+     */
+    public function __construct(array $directoryPaths)
```
crisu83: Why not call this `$filePaths` too?

Jalle19: Leftover from refactoring

src/Language/MultiFileSourceBuilder.php
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * @throws InvariantException
+     */
+    public function buildSource(): Source
+    {
+        $combinedSource = '';
+
+        foreach ($this->filePaths as $filePath) {
+            if (!\file_exists($filePath) || !\is_readable($filePath)) {
+                throw new InvariantException(sprintf('The file %s cannot be found or is not readable', $filePath));
```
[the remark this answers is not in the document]
Jalle19: Same what?

tests/Functional/Language/FileSourceBuilderTest.php
Jalle19: Different builders need different arguments, I don't think we can make those a part of the interface?

crisu83: Oh right, let's keep it as is then.

src/Language/MultiFileSourceBuilder.php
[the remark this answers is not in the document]
crisu83: The use of `InvariantException` here.

src/Language/FileSourceBuilder.php
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * @inheritdoc
+     * 
+     * @throws InvariantException
+     */
+    public function buildSource(): Source
+    {
+        if (!\file_exists($this->filePath) || !\is_readable($this->filePath)) {
+            throw new InvariantException(sprintf('The file %s cannot be found or is not readable', $this->filePath));
```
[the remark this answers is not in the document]
Jalle19: We could throw RuntimeException or I could introduce a new FileNotFoundException?

Jalle19:
Addressed all the comments and updated the README

PR #270, opened 2018-08-20
Fix Scrutinizer issues

crisu83:
Fix some of the issues noticed by Scrutinizer.


Jalle19:
That went well :)

crisu83:
@Jalle19 So far I've fixed 91 issues. I'll continue later tonight.

Jalle19:
You're right, GitHub reported the results as "7 new issues found" for some reason - nothing about what got fixed

crisu83:
I ended up fixing 110 issues, without touching the `Execution` namespace - We can do that later. If everything seems fine we can merge this.

tests/Functional/Execution/SchemaTest.php
```diff
@@ -279,9 +279,9 @@ public function testExecuteUsingASchemaWithCustomScalarType()
         $blogArticle = newObjectType([
             'name'   => 'Article',
             'fields' => [
-                'id'        => ['type' => newNonNull(String())],
-                'title'     => ['type' => String()],
-                'body'      => ['type' => String()],
+                'id'        => ['type' => newNonNull(StringType())],
```
hungneox: @crisu83 Why don't we also have `NewNonNull()` function? Why capitalize `StringType()` but not `NewNonNull()`? Just a small thing but it make my eyes itchy, because I think it's public API and it's not consistency. Is there any change that we change it later 😵

tests/Functional/Execution/VariablesTest.php
```diff
@@ -1259,9 +1259,9 @@ public function testCustomDateTimeScalarType()
         $TestInputObject = newInputObjectType([
             'name'   => 'TestInputObject',
             'fields' => [
-                'a' => ['type' => String()],
-                'b' => ['type' => newList(String())],
-                'c' => ['type' => newNonNull(String())],
+                'a' => ['type' => StringType()],
+                'b' => ['type' => newList(StringType())],
```
hungneox: All so for `newList`

tests/Functional/Execution/SchemaTest.php
crisu83: You're correct, I'll change these types to be non-capitalized.

src/Language/blockStringValue.php
```diff
@@ -12,16 +12,16 @@
...
+    $lines        = false === $lines ? [] : $lines;
+    $lineCount    = \count($lines);
     $commonIndent = null;
 
     for ($i = 1; $i < $lineCount; $i++) {
         $line   = $lines[$i];
         $indent = leadingWhitespace($line);
 
-        if ($indent < mb_strlen($line) && ($commonIndent === null || $indent < $commonIndent)) {
+        if ($indent < \mb_strlen($line) && ($commonIndent === null || $indent < $commonIndent)) {
```
hungneox: Not a big deal but should we change to `null === $commonIndent` for consistency?

tests/Functional/Execution/SchemaTest.php
hungneox: Okay 🎖

src/Language/blockStringValue.php
crisu83: Fixed this as well.

PR #273, opened 2018-08-21
Named type node interface

crisu83:
Add interface for named type nodes.

### DIFF
--- a/src/Language/Node/EnumTypeDefinitionNode.php
+++ b/src/Language/Node/EnumTypeDefinitionNode.php
@@ -5,7 +5,7 @@ namespace Digia\GraphQL\Language\Node;
 use Digia\GraphQL\Language\Location;
 
 class EnumTypeDefinitionNode extends AbstractNode implements TypeSystemDefinitionNodeInterface, DirectivesAwareInterface,
-    NameAwareInterface
+    NameAwareInterface, NamedTypeNodeInterface
 {
     use DescriptionTrait;
     use NameTrait;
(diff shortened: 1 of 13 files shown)


src/Language/Node/EnumTypeDefinitionNode.php
```diff
@@ -5,7 +5,7 @@
 use Digia\GraphQL\Language\Location;
 
 class EnumTypeDefinitionNode extends AbstractNode implements TypeSystemDefinitionNodeInterface, DirectivesAwareInterface,
-    NameAwareInterface
+    NameAwareInterface, NamedTypeNodeInterface
```
Jalle19: > Lists of implements MAY be split across multiple lines, where each
subsequent line is indented once. When doing so, the first item in the list
MUST be on the next line, and there MUST be only one interface per line

src/Language/Node/InputObjectTypeDefinitionNode.php
```diff
@@ -5,7 +5,7 @@
 use Digia\GraphQL\Language\Location;
 
 class InputObjectTypeDefinitionNode extends AbstractNode implements TypeSystemDefinitionNodeInterface,
-    DirectivesAwareInterface, NameAwareInterface
+    DirectivesAwareInterface, NameAwareInterface, NamedTypeNodeInterface
```
Jalle19: Same here and in other classes that have implements on multiple lines

src/Language/Node/EnumTypeDefinitionNode.php
crisu83: There are literally tens of files where the interfaces has been split like this. I'm not going to change them all in this PR. We can do this later.

PR #280, opened 2018-09-18
Fix coercing false non-null booleans

### LINKED ISSUE #279: Cannot pass false as the value to a "Boolean!" argument
### What did you do?

Tried a query like this:

```graphql
mutation changeCustomerNumber($customerId: ID!, $customerNumber: String!, $postalCode: String!, $isPersonal: Boolean!) {
  changeCustomerNumber(customerId: $customerId, customerNumber: $customerNumber, postalCode: $postalCode, isPersonal: $isPersonal)
}
```

With these variables:

```json
{
  "customerId": "blabla",
  "customerNumber": "123",
  "isPersonal": false,
  "postalCode": "12345"
}
```
### What did you expect to happen?

Everything to work

### What actually happened?

```json
{
  "errors": [
    {
      "message": "Variable \"$isPersonal\" got invalid value false; Expected non-nullable type Boolean! not to be null.",
      "locations": [
        {
          "line": 1,
          "column": 97
        }
      ],
      "path": []
    }
  ],
  "data": null
}
```

Jalle19:
Fixes #279

### DIFF
--- a/src/Execution/ValuesHelper.php
+++ b/src/Execution/ValuesHelper.php
@@ -293,7 +293,7 @@ class ValuesHelper
         NodeInterface $blameNode,
         ?Path $path
     ): CoercedValue {
-        if (empty($value)) {
+        if (null === $value) {
             return new CoercedValue(null, [
                 $this->buildCoerceException(
                     sprintf('Expected non-nullable type %s not to be null', (string)$type),
--- a/tests/Functional/Execution/ValuesHelperTest.php
+++ b/tests/Functional/Execution/ValuesHelperTest.php
@@ -10,7 +10,7 @@ use function Digia\GraphQL\Type\newObjectType;
 use function Digia\GraphQL\Type\newSchema;
 use function Digia\GraphQL\Type\stringType;
 
-class ValueHelperTest extends TestCase
+class ValuesHelperTest extends TestCase
 {
     /**
      * @throws \Digia\GraphQL\Error\InvariantException
@@ -34,7 +34,6 @@ class ValueHelperTest extends TestCase
                     ]
                 ])
         ]);
-
 
         $documentNode = parse('query Hello($name: String) { Greeting(name: $name) }');
         $operation  = $documentNode->getDefinitions()[0];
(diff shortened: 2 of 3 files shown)


src/Execution/ValuesHelper.php
```diff
@@ -293,7 +293,7 @@ protected function coerceValueForNonNullType(
         NodeInterface $blameNode,
         ?Path $path
     ): CoercedValue {
-        if (empty($value)) {
+        if ($value === null) {
```
crisu83: @Jalle19 Can you flip this comparison for consistency?

tests/Functional/Execution/ValuesHelperTest.php
```diff
@@ -0,0 +1,102 @@
...
+                                'name' => [
+                                    'type' => stringType(),
+                                ]
+                            ]
+                        ]
+                    ]
+                ])
+        ]);
+
+
```
crisu83: Also here's an extra line-break.

tests/Functional/Execution/ValuesHelperTest.php
```diff
@@ -0,0 +1,102 @@
...
+     * @throws \Digia\GraphQL\Error\InvariantException
+     * @throws \Digia\GraphQL\Error\SyntaxErrorException
+     */
+    public function testCoerceVariableValues(): void
+    {
+        // Test that non-nullable booleans are handled correctly
+        $schema = newSchema([
+            'query' =>
+                newObjectType([
+                    'name'   => 'NonNullBoolean',
```
crisu83: This type name should start with a lower-case letter.

tests/Functional/Execution/ValuesHelperTest.php
```diff
@@ -0,0 +1,106 @@
...
+                                'name' => [
+                                    'type' => stringType(),
+                                ]
+                            ]
+                        ]
+                    ]
+                ])
+        ]);
+
+        $documentNode = parse('query Hello($name: String) { Greeting(name: $name) }');
```
crisu83: Ah I missed this query last time, I would use line-breaks here too.

Jalle19: Like earlier, this was already like this, GitHub's diff just doesn't understand that this is existing code

PR #281, opened 2018-09-19
Refactor VisitorInterface to allow for more fine-grained control

### LINKED ISSUE #42: Fix testMaintainsTypeInfoDuringEdit

Jalle19:
This started out as an attempt to fix #42 . Currently some tests are failing because "break" is not implemented.

### DIFF
--- a/src/Language/Node/AbstractNode.php
+++ b/src/Language/Node/AbstractNode.php
@@ -5,7 +5,9 @@ namespace Digia\GraphQL\Language\Node;
 use Digia\GraphQL\GraphQL;
 use Digia\GraphQL\Language\Location;
 use Digia\GraphQL\Language\NodeBuilderInterface;
+use Digia\GraphQL\Language\Visitor\VisitorBreak;
 use Digia\GraphQL\Language\Visitor\VisitorInterface;
+use Digia\GraphQL\Language\Visitor\VisitorResult;
 use Digia\GraphQL\Util\ArrayToJsonTrait;
 use Digia\GraphQL\Util\SerializationInterface;
 
@@ -126,12 +128,7 @@ abstract class AbstractNode implements NodeInterface, SerializationInterface
     }
 
     /**
-     * @param VisitorInterface   $visitor
-     * @param mixed              $key
-     * @param NodeInterface|null $parent
-     * @param string[]           $path
-     * @param NodeInterface[]    $ancestors
-     * @return NodeInterface|null
+     * @inheritdoc
      */
     public function acceptVisitor(
         VisitorInterface $visitor,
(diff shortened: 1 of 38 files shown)


Jalle19:
Restored breaking functionality

src/Language/Visitor/AbstractReturnValue.php
```diff
@@ -0,0 +1,61 @@
...
+
+use Digia\GraphQL\Language\Node\NodeInterface;
+
+abstract class AbstractReturnValue
+{
+
+    public const ACTION_NO_ACTION = 'NO_ACTION';
+    public const ACTION_BREAK     = 'BREAK';
+    public const ACTION_DELETE    = 'DELETE';
+    public const ACTION_REPLACE   = 'REPLACE';
```
crisu83: I would consider using an enumerable class for this, e.g. `VisitorActionEnum`.

src/Language/Visitor/EnterReturnValue.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Digia\GraphQL\Language\Visitor;
+
+final class EnterReturnValue extends AbstractReturnValue
+{
+
+    public const ACTION_SKIP = 'SKIP';
```
crisu83: What is this used for?

src/Language/Visitor/EnterReturnValue.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Digia\GraphQL\Language\Visitor;
+
+final class EnterReturnValue extends AbstractReturnValue
```
crisu83: Couldn't we just get by with one class, e.g. `VisitorResult`? If not, then I'd call these two classes `VisitorEnterResult` and `VisitorLeaveResult`.

crisu83:
@Jalle19 This is not WIP anymore, right? If so, please remove the prefix from the title.

src/Language/Visitor/EnterReturnValue.php
Jalle19: They're not used anywhere, I just copied the values from the reference implementation

src/Language/Visitor/EnterReturnValue.php
Jalle19: I suppose we could, it's just that enter has one more possible value than leave

crisu83: Ok fair enough, let's keep it like this then. 👍

crisu83:
@Jalle19 Feel free to merge this. Thanks a lot for taking the time to refactor the visitor logic. 👍

Jalle19:
I'm still thinking we could combine the two return value classes, and skip the unused constants

crisu83:
What’s up w this PR @Jalle19?

crisu83:
@Jalle19 I say we just merge this and do the rest of the refactoring when you have time.

Jalle19:
I was actually gonna look at this today, so don't merge quite yet

Jalle19:
@crisu83 updated

PR #283, opened 2018-09-19
String source builder

Jalle19:
String source builder, stop passing around both strings and Source objects.

This way we can use static type-hinting, which in this case unearthed a bug with how lex methods are called in the parser.


Jalle19:
I wouldn't consider this an API breakage since the most high-level way of using the library (api.php) remains unchanged.

PR #284, opened 2018-09-19
Small improvement for defaultTypeResolver

hungneox:
- Use value __typename as name of the resolved type
- Port two tests from reference implementation


hungneox:
@crisu83 I have fixed those issues based on your comments.

tests/Functional/Schema/SchemaBuilderTest.php
```diff
@@ -0,0 +1,158 @@
+<?php
+/**
```
crisu83: @hungneox Could you remove the author block for consistency.

hungneox: @crisu83 I merged it already :< I will remove this on the next PR. 😢

PR #288, opened 2018-10-01
Error handler

crisu83:
Add initial support for error handling. This will allow developers to swap out the default error handler in the DI container for their own in order to add metadata to their errors based on which exception was originally thrown. 

The specification for error response format can be found here: https://github.com/facebook/graphql/blob/master/spec/Section%207%20--%20Response.md


src/Execution/Executor.php
```diff
@@ -104,9 +106,9 @@ public function execute(): ?array
                 ? $this->executeFieldsSerially($objectType, $rootValue, $path, $fields)
                 : $this->executeFields($objectType, $rootValue, $path, $fields);
         } catch (\Throwable $ex) {
-            $this->context->addError(new ExecutionException($ex->getMessage()));
-
-            return [null];
+            $error = new ExecutionException($ex->getMessage(), $fields, null, null, null, null, $ex);
```
hungneox: I just wonder is it worth to create a temp variable here?

src/Error/GraphQLException.php
```diff
@@ -96,6 +103,7 @@ public function __construct(
         $this->resolveLocations($positions, $source);
 
         $this->path              = $path;
+        $this->extensions        = $extensions;
```
hungneox: @crisu83 What do we use extensions for?

src/Execution/Executor.php
crisu83: It's not temporary, it's used twice.

src/Error/GraphQLException.php
crisu83: It's in the spec, you can add error codes, or whatever else you like there.

src/Execution/Executor.php
hungneox: @crisu83 In this block, you only add it to handleError?

crisu83: And to add it to the context.

src/Execution/Executor.php
```diff
@@ -946,6 +950,23 @@ protected function createResolveInfo(
         );
     }
 
+    /**
+     * @param ExecutionException $error
+     */
+    protected function handleError(ExecutionException $error)
+    {
+        $this->getErrorHandler()->handleError($error);
```
hungneox: @crisu83 I see, but you add them here. For the block above I think this works fine
```
$this->handleError(new ExecutionException($ex->getMessage(), $fields, null, null, null, null, $ex));
```

src/Execution/Executor.php
crisu83: Ah true, I moved that into the `handleError` method.

src/Execution/Executor.php
crisu83: Yeah I agree, we can change this.

crisu83: Yeah I agree, we can change this.

crisu83: Yeah I agree, we can change this.

crisu83: Yeah I agree, we can change this.

src/Error/GraphQLException.php
```diff
@@ -181,6 +207,26 @@ public function getOriginalErrorMessage(): ?string
...
 
+    /**
+     * @inheritdoc
+     */
+    public function toArray(): array
+    {
+        return [
+            'message'   => $this->message,
+            'locations' => $this->getLocationsAsArray(),
+            'path'      => $this->path,
```
Jalle19: Extensions should be included here when they're not null, right?

src/Execution/Executor.php
```diff
@@ -946,6 +950,23 @@ protected function createResolveInfo(
...
+        $this->context->addError($error);
+    }
+
+    /**
+     * @return ErrorHandlerInterface
+     */
+    protected function getErrorHandler(): ErrorHandlerInterface
+    {
+        return GraphQL::make(ErrorHandlerInterface::class);
```
Jalle19: Wouldn't it be better to inject the error handler through the constructor?

crisu83: Hmm yes, you're right.

src/Error/GraphQLException.php
crisu83: Yes, I'll add it there.

crisu83: Done.

src/Execution/Executor.php
crisu83: Done.

src/Execution/Executor.php
crisu83: Done.

src/Execution/Executor.php
crisu83: Done.

src/Error/ErrorHandlerInterface.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Digia\GraphQL\Error;
+
+interface ErrorHandlerInterface
+{
+    /**
+     * @param ExecutionException $error
+     */
+    public function handleError(ExecutionException $error);
```
Jalle19: Here it's called `$error`, in the implementation it's called `$exception`

src/Error/ErrorProvider.php
```diff
@@ -0,0 +1,23 @@
...
+    protected $provides = [
+        ErrorHandlerInterface::class,
+    ];
+
+    /**
+     * @inheritdoc
+     */
+    public function register()
+    {
+        $this->container->add(ErrorHandlerInterface::class, ErrorHandler::class, true/* $shared */);
```
Jalle19: Since you merged the league/container update PR this needs updating too I guess?

src/Error/GraphQLException.php
Jalle19: Where did you add it?

src/Execution/Executor.php
hungneox: @crisu83 Why not just

```
$this->handleError(new ExecutionException(...
```

src/Error/ErrorHandlerInterface.php
crisu83: Good catch, I'll fix it.

src/Error/ErrorProvider.php
crisu83: True, I'll change this.

src/Execution/Execution.php
```diff
@@ -141,6 +156,6 @@ protected function createContext(
      */
     protected function createExecutor(ExecutionContext $context): Executor
     {
-        return new Executor($context, new FieldCollector($context));
+        return new Executor($context, new FieldCollector($context), $this->errorHandler);
```
Jalle19: Should we consider refactoring this so we don't have to pass the error handler through this `Execution` thing just to get it into the `Executor`?

src/Execution/Executor.php
```diff
@@ -946,6 +957,15 @@ protected function createResolveInfo(
         );
     }
 
+    /**
+     * @param ExecutionException $error
+     */
+    protected function handleError(ExecutionException $error)
```
Jalle19: `: void`

src/Error/GraphQLException.php
```diff
@@ -181,6 +207,32 @@ public function getOriginalErrorMessage(): ?string
...
+     */
+    public function toArray(): array
+    {
+        $result = [
+            'message'    => $this->message,
+            'locations'  => $this->getLocationsAsArray(),
+            'path'       => $this->path,
+        ];
+
+        if (!empty($this->extensions)) {
```
Jalle19: This should probably be `$this->extensions !== null`, an implementator might want extensions to be included even if it's an empty array

Jalle19: Either that or extensions should always be an array, not `?array`

src/Execution/Executor.php
crisu83: We don't use that anywhere yet, we should though, but I'll keep it like this for consistency.

src/Error/GraphQLException.php
crisu83: True, I'll fix this.

src/Execution/Executor.php
Jalle19: It's used in the same class

src/Error/GraphQLException.php
```diff
@@ -181,6 +207,32 @@ public function getOriginalErrorMessage(): ?string
...
+    public function toArray(): array
+    {
+        $result = [
+            'message'    => $this->message,
+            'locations'  => $this->getLocationsAsArray(),
+            'path'       => $this->path,
+        ];
+
+        if (null !== $this->extensions) {
+            $result['extensions'] = $this->extensions;
```
Jalle19: Can you update the tests for this method to check that this extensions handling works correctly?

src/Execution/Execution.php
crisu83: That might be a bit troublesome because we would need a factory for `Executor` in order to provide both injected and provided arguments to its constructor.

src/Error/GraphQLException.php
crisu83: I added tests for this and some other things.

src/Execution/Executor.php
crisu83: Where? I can't see it anywhere.

PR #294, opened 2018-10-03
Refactor error handler injection

crisu83:
I decided to refactor how the error handler is injected in order to ease the use of an error handler. Before this change developers had to manually inject their error handler into the container, which is far from developer-friendly. After this change, the error handler can be provided directly to the executor as an argument at execution time. Also, there is no default error handler (the default error handler did nothing before) and the error handler can also be a lambda function in simple use cases.


crisu83:
Well sure I'll add unit tests for the error handler itself.

PR #295, opened 2018-10-03
Use GraphQLException for caught exceptions during execution 

hungneox:
- Fix some typos
- Change ExecutionException to GraphQLException for caught exception


hungneox:
@Jalle19 Sorry, I will close this PR, I had dicussion with @crisu83. It doesn't solve the issue. Basically we shouldn't wrap client exception into our ExecutionException, since it's not really ExecutionException.

PR #297, opened 2018-10-08
Fix handling of errors in mutations

crisu83:


src/Execution/Executor.php
```diff
@@ -114,8 +114,8 @@ public function execute(): ?array
             $result = $operation->getOperation() === 'mutation'
                 ? $this->executeFieldsSerially($objectType, $rootValue, $path, $fields)
                 : $this->executeFields($objectType, $rootValue, $path, $fields);
-        } catch (\Throwable $ex) {
-            $this->handleError(new ExecutionException($ex->getMessage(), $fields, null, null, null, null, $ex));
+        } catch (ExecutionException $ex) {
```
Jalle19: Now \Throwable and other exceptions won't ever be caught?

crisu83: This should be correct, `\Throwable`s are already elsewhere wrapped after `ExecutionException`s.

crisu83: What we can do is add another `catch` clause, just to be sure though. Good point.

src/Execution/Executor.php
```diff
@@ -217,8 +218,8 @@ public function executeFieldsSerially(
 
         $promise->then(function ($resolvedResults) use (&$finalResults) {
             $finalResults = $resolvedResults ?? [];
-        })->otherwise(function ($ex) {
-            $this->context->addError($ex);
+        })->otherwise(function (ExecutionException $ex) {
```
hungneox: What if `$ex` is not `ExecutionException`? Do we lose that exception?

src/Execution/Executor.php
```diff
@@ -379,7 +385,7 @@ protected function executeFields(
                 continue;
             }
 
-            $doesContainPromise = $doesContainPromise || $this->isPromise($result);
+            $doesContainPromise = $doesContainPromise || $result instanceof ExtendedPromiseInterface;
```
hungneox: @crisu83 Not sure why you don't like the `isPromise` method, but we use it in ~3 places at least.

src/Execution/Executor.php
Jalle19: Sorry, clicked the wrong button. I'm also wondering if this is correct?

src/Execution/Executor.php
crisu83: Because it’s better to use `instanceof` instead of the `@var` comments.

hungneox: @crisu83 If you insist, feel free to change, but we actually don't really use it in the code.

src/Execution/Executor.php
crisu83: You're right, I've fixed this now.

src/Execution/Executor.php
```diff
@@ -115,7 +114,7 @@ public function execute(): ?array
                 ? $this->executeFieldsSerially($objectType, $rootValue, $path, $fields)
                 : $this->executeFields($objectType, $rootValue, $path, $fields);
         } catch (\Throwable $ex) {
-            $this->handleError(new ExecutionException($ex->getMessage(), $fields, null, null, null, null, $ex));
+            $this->handleError(new ExecutionException($ex->getMessage(), null, null, null, null, null, $ex));
```
hungneox: @crisu83 not sure why you removed the `$fields`?

PR #299, opened 2018-10-08
Resolver middleware

crisu83:
Add support for resolver middleware.

This feature was inspired by [graphql-middlware](https://github.com/prisma/graphql-middleware).


src/Schema/Resolver/ResolverCollection.php
```diff
@@ -0,0 +1,82 @@
...
+    public function getResolveCallback(): ?callable
+    {
+        return function ($fieldName) {
+            $resolver = $this->getResolver($fieldName);
+
+            if ($resolver instanceof ResolverInterface) {
+                return $resolver->getResolveCallback();
+            }
+
+            if ($resolver instanceof \Closure) {
```
Jalle19: `\is_callable`

crisu83: This is actually correct.

PR #300, opened 2018-10-08
Named resolver middleware

crisu83:
Add support for applying specific middleware to resolvers.

### DIFF
--- a/src/Schema/Resolver/ResolverRegistry.php
+++ b/src/Schema/Resolver/ResolverRegistry.php
@@ -44,7 +44,7 @@ class ResolverRegistry implements ResolverRegistryInterface
         $resolver = $this->getResolver($typeName);
 
         $resolver = $resolver instanceof ResolverCollection
-            ? $resolver->getResolveCallback()($fieldName)
+            ? $resolver->getResolver($fieldName)
             : $resolver;
 
         $resolveCallback = $resolver instanceof ResolverInterface
@@ -56,7 +56,11 @@ class ResolverRegistry implements ResolverRegistryInterface
         }
 
         if (null !== $this->middleware) {
-            return $this->applyMiddleware($resolveCallback, \array_reverse($this->middleware));
+            $middleware = $resolver instanceof ResolverInterface
+                ? $this->getMiddlewareToApply($resolver, $this->middleware)
+                : $this->middleware;
+
+            return $this->applyMiddleware($resolveCallback, \array_reverse($middleware));
         }
 
         return $resolveCallback;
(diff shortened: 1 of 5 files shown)


tests/Unit/Schema/Resolver/ResolverRegistryTest.php
```diff
@@ -205,3 +232,13 @@ public function resolve($rootValue, array $arguments, $context = null, ?ResolveI
         return \sprintf('Hello %s!', $arguments['name'] ?? 'world');
     }
 }
+
+class HelloResolverWithSpecificMiddleware extends HelloResolver
+{
+    public function getMiddleware(): ?array
+    {
+        return [
+            'logInput',
```
crisu83: We could use the FQCN as well.

src/Schema/Resolver/ResolverRegistry.php
```diff
@@ -44,7 +44,7 @@ public function getFieldResolver(string $typeName, string $fieldName): ?callable
         $resolver = $this->getResolver($typeName);
 
         $resolver = $resolver instanceof ResolverCollection
-            ? $resolver->getResolveCallback()($fieldName)
+            ? $resolver->getResolver($fieldName)
```
crisu83: This was actually a bug that I found while working on this.

tests/Unit/Schema/Resolver/ResolverRegistryTest.php
spawnia: I personally prefer FQCN over stringly typed short names.

crisu83: Agreed, this is very JavaScripty now that I look at it.

crisu83: I’ll change this in a moment.

spawnia: Same when registering them. `SomeClass::class` is the safest option, and even easier to type if you have auto-completion

crisu83: @spawnia I would prefer using the `new` keyword for registering them because this allows you to do DI if necessary. Or would you prefer the option of using both the `new` keyword and FQCN for both resolvers and middleware?

spawnia: The way you put it now is perfect. Eliminating the need to come up with a short name for them is better.

src/Schema/Resolver/ResolverRegistry.php
hungneox: 👍

PR #303, opened 2018-10-15
Fix coercing of false values

crisu83:
Fixed a bug that caused `false` values to be coerced as `null`.

### DIFF
--- a/src/Execution/ValuesHelper.php
+++ b/src/Execution/ValuesHelper.php
@@ -258,8 +258,8 @@ class ValuesHelper
             return $this->coerceValueForNonNullType($value, $type, $blameNode, $path);
         }
 
-        if (empty($value)) {
-            return new CoercedValue(null, null);
+        if (null === $value) {
+            return new CoercedValue(null);
         }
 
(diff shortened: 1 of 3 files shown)


src/Execution/ValuesHelper.php
```diff
@@ -259,8 +258,8 @@ private function coerceValue($value, $type, $blameNode, ?Path $path = null): Coe
             return $this->coerceValueForNonNullType($value, $type, $blameNode, $path);
         }
 
-        if (empty($value)) {
-            return new CoercedValue(null, null);
+        if (null === $value) {
```
hungneox: 👍

hungneox: I thought we fix this already :-?

crisu83: Apparently we didn't or maybe it got unfixed at some point. We did have unit tests for this, but `assertEquals` apparently does a loose comparison of array values so the tests didn't fail even though they should have.

hungneox: OK, got it.

PR #311, opened 2018-10-22
Fix "Cannot use false as value for non-null booleans in input types"

### LINKED ISSUE #310: Cannot use false as value for non-null booleans in input types
Similar to https://github.com/digiaonline/graphql-php/issues/279

Jalle19:
Fixes #310


Jalle19:
Let's wait until GitHub's service hooks work again before merging this

PR #318, opened 2018-10-24
Error handler middleware

crisu83:
Add support for error handler middleware.


Jalle19:
I pushed some minor cleanups. @crisu83 what should we do about errors that occur before the main execution phase?

* schema validation errors
* query parsing errors
* query validation errors
* createContext() errors

crisu83:
@Jalle19 I guess we could handle those in the same manner with some minor refactoring. We could actually make this a non-breaking change by some additional refactoring. We just need a `CallableMiddleware` that is used to wrap any `callable` that is passed in as the error handler.

src/GraphQL.php
```diff
@@ -249,18 +252,82 @@ public static function execute(
...
+     * @param null                       $contextValue
+     * @param array                      $variableValues
+     * @param null|string                $operationName
+     * @param callable|null              $fieldResolver
+     * @param ErrorHandlerInterface|null $errorHandler
+     * @return ExecutionResult
+     * @throws InvariantException
+     * @throws SyntaxErrorException
+     */
+    public static function process(
```
Jalle19: Why move this to a static method?

crisu83: This wasn't a part of the `GraphQL` class at all before, only the public API.

src/api.php
```diff
@@ -162,44 +164,33 @@ function printNode(NodeInterface $node): string
...
-        return (new ExecutionResult([], [$error]))->toArray();
-    }
-
-    $validationErrors = validate($schema, $document);
-    if (!empty($validationErrors)) {
-        return (new ExecutionResult([], $validationErrors))->toArray();
-    }
-
-    $result = execute(
+    $result = GraphQL::process(
```
Jalle19: This now doesn't go through the container, it used to do that.

Jalle19: Actually never mind, I read this wrong

Jalle19:
@crisu83 this should be good to go, it's up to you if you want to fix the null check

Jalle19:
Oh yeah and please update the documentation

src/Execution/Execution.php
```diff
@@ -58,9 +58,9 @@ public function execute(
             });
         }
 
-        if (null !== $errorHandler) {
+        if ($errorHandler instanceof ErrorHandlerInterface) {
```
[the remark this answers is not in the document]
Jalle19: Pushed some minor changes

PR #321, opened 2018-10-31
Execution strategies

crisu83:
Add initial support for execution strategies.


crisu83:
@benjamindulau can you please try your use-case with this branch just to see if there is any difference. If it's still doing all those method calls I'll start investigating where the problem lies. Thank you in advance.

benjamindulau:
@crisu83 I'll try this as soon as I can! (by the end of the week). Thanks ;-)

src/Execution/Strategy/ParallelExecutionStrategy.php
```diff
@@ -0,0 +1,50 @@
...
+ */
+class ParallelExecutionStrategy extends AbstractExecutionStrategy
+{
+    /**
+     * @inheritdoc
+     */
+    public function executeFields(ObjectType $parentType, $rootValue, array $path, array $fields)
+    {
+        $results            = [];
+        $doesContainPromise = false;
```
Jalle19: Could be just `$containsPromise`

src/Util/utils.php
```diff
@@ -158,6 +158,43 @@ function keyValueMap(array $array, callable $keyFn, callable $valFn): array
     }, []);
 }
 
+/**
+ * @param array $map
+ * @return PromiseInterface
+ */
+function promiseForMap(array $map): PromiseInterface
```
Jalle19: IMO we need tests for this one and `promiseReduce`

src/Execution/Execution.php
```diff
@@ -52,10 +47,57 @@ public function execute(
...
+            $result = $strategy->execute();
+        } catch (ExecutionException $exception) {
+            $context->addError($exception);
+        } catch (\Throwable $exception) {
+            $context->addError(
+                new ExecutionException($exception->getMessage(), null, null, null, null, null, $exception)
+            );
+        }
+
+        if ($result instanceof PromiseInterface) {
```
Jalle19: In https://sergeyzhuk.me/2018/01/16/reactphp-managing-promises/ they mention "I don’t know exactly what the resolver will give me", i.e. it seems it's easy to wrap any result in a promise. This could potentially simplify the execution logic a bit, don't you think?

src/Execution/Strategy/ParallelExecutionStrategy.php
crisu83: Yeah I agree.

src/Execution/Execution.php
crisu83: It could, but I don't how that would affect the performance though.

Jalle19: I doubt it affects performance

crisu83: @Jalle19 Do you want to give it a try?

Jalle19: What do you mean?

crisu83: @Jalle19 Changing the code to always use promises.

Jalle19:
@crisu83 can you:

1) Drop all the ValuesHelper stuff, it was recently refactored anyway
2) Drop all the unrelated formatting/noinspection changes?

crisu83:
@Jalle19 1. is easy, not sure how easy it is to undo 2.

src/Execution/Execution.php
```diff
@@ -3,34 +3,29 @@
...
     public function execute(
         Schema $schema,
         DocumentNode $documentNode,
         $rootValue = null,
         $contextValue = null,
         array $variableValues = [],
-        ?string $operationName = null,
-        ?callable $fieldResolver = null,
+        string $operationName = null,
+        callable $fieldResolver = null,
```
Jalle19: These should not have been changed, if you can accept null then it was to be `?foo`

Jalle19:
@crisu83 just `git reset HEAD^`, commit only the things you want to change

crisu83:
@Jalle19 1. Is now done.

src/Execution/Execution.php
crisu83: This is now fixed.

crisu83:
@Jalle19 is too many changes for that.

Jalle19:
If it is then you forgot to push

crisu83:
It says "Outdated" above your comment.

crisu83:
Well, I renamed it because in other implementations they call it `ValuesResolver`, not helper.

Jalle19:
@crisu83 would have been good to do it in a separate PR

crisu83:
@Jalle19 but I take it that you want to call it `ValuesHelper`? It's just a name change, nothing more. Also all the tests pass, so it should be fine as is.

tests/Functional/Execution/AbstractPromiseTest.php
```diff
@@ -22,19 +22,18 @@ class AbstractPromiseTest extends TestCase
 
     /**
      * isTypeOf used to resolve runtime type for Interface
-     *
-     * @throws \Digia\GraphQL\Error\InvariantException
-     * @throws \Digia\GraphQL\Error\SyntaxErrorException
      */
     public function testIsTypeOfUsedToResolveFunctionForInterface()
     {
+        /** @noinspection PhpUnhandledExceptionInspection */
```
Jalle19: I'm not a fan of these, why not just keep the throws tags?

tests/Functional/Execution/ExecutionTest.php
```diff
@@ -888,42 +887,44 @@ public function testFullResponsePathIsIncludedForNonNullableFields()
...
+        $this->assertSame([
             'errors' => [
                 [
                     'message'   => 'Catch me if you can!',
                     'locations' => [
                         [
-                            'line'   => 7,
-                            'column' => 21
-                        ]
+                            'line'   => 6,
```
Jalle19: Why did this suddenly change?

tests/Functional/Execution/ExecutionTest.php
```diff
@@ -1462,40 +1474,38 @@ public function testUsesACustomFieldResolver()
...
+//                [
+//
+//                    'message'   => 'Cannot return null for non-nullable field Greeting.hello.',
+//                    'locations' => null,
+//                    'path'      => null,
+//                ]
+//            ],
+//            'data'   => null
+//        ], $executionResult);
+//    }
```
Jalle19: What happened to this test?

tests/Functional/starWarsSchema.php
```diff
@@ -22,21 +22,26 @@ function starWarsSchema()
...
-            'friends' => function ($droid) {
+        'Droid'     => [
+            'friends'         => function ($droid) {
                 return getFriends($droid);
             },
             'secretBackstory' => function () {
                 throw new \Exception('secretBackstory is secret.');
             },
         ],
+        'Character' => [
```
Jalle19: This wasn't here before, should it be?

Jalle19:
> @Jalle19 I take it that you want to call it `ValuesHelper`? It's just a name change, nothing more. Also, all the tests pass, so it should be fine as is.

I don't really care what it's called, I just don't like the fact that this PR has a bunch of unrelated changes

tests/Functional/Execution/AbstractPromiseTest.php
crisu83: It's a test, they shouldn't throw anything.

tests/Functional/Execution/ExecutionTest.php
crisu83: Because I fixed the query string.

tests/Functional/Execution/ExecutionTest.php
crisu83: It's redundant if I'm not mistaken.

crisu83: Nevermind, my mistake. I re-added the test case now. Good catch.

tests/Functional/starWarsSchema.php
crisu83: Yes, because I removed the automatic resolving using the `type` field as it's not part of the spec.

crisu83:
@Jalle19 Should we merge this?

PR #347, opened 2019-07-25
Fix definition printer tests

crisu83:
Fix the ignored test cases in `DefinitionPrinterTest`.

### DIFF
--- a/src/Language/DirectiveLocationEnum.php
+++ b/src/Language/DirectiveLocationEnum.php
@@ -13,6 +13,7 @@ class DirectiveLocationEnum
     public const FRAGMENT_DEFINITION = 'FRAGMENT_DEFINITION';
     public const FRAGMENT_SPREAD     = 'FRAGMENT_SPREAD';
     public const INLINE_FRAGMENT     = 'INLINE_FRAGMENT';
+    public const VARIABLE_DEFINITION = 'VARIABLE_DEFINITION';
     // Type System Definitions
     public const SCHEMA                 = 'SCHEMA';
     public const SCALAR                 = 'SCALAR';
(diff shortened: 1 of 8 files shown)


crisu83:
All associated tests pass now and phpstan doesn't raise any new errors, so this should be good to merge. @Jalle19 please take a look and tell me if there is anything I may have overlooked. Thanks in advance.

src/Type/IntrospectionProvider.php
```diff
@@ -215,14 +218,13 @@ protected function registerIntrospectionTypes()
                     'name'            => GraphQL::TYPE_INTROSPECTION,
                     'isIntrospection' => true,
                     'description'     =>
-                        'The fundamental unit of any GraphQL Schema is the type. There are ' .
-                        'many kinds of types in GraphQL as represented by the `__TypeKind` enum.' .
-                        '\n\nDepending on the kind of a type, certain fields describe ' .
```
crisu83: Took a while to find that this line was not in double-quotes, so linebreaks didn't work as expected. 😐

src/Language/DirectiveLocationEnum.php
```diff
@@ -13,6 +13,7 @@ class DirectiveLocationEnum
     public const FRAGMENT_DEFINITION = 'FRAGMENT_DEFINITION';
     public const FRAGMENT_SPREAD     = 'FRAGMENT_SPREAD';
     public const INLINE_FRAGMENT     = 'INLINE_FRAGMENT';
+    public const VARIABLE_DEFINITION = 'VARIABLE_DEFINITION';
```
crisu83: This location was missing. Maybe it was added to the reference implementation afterwards? 🤔

PR #349, opened 2019-07-26
Fix null variables and default values

### LINKED ISSUE #239: Fix ambiguity with null variable values and default values

crisu83:
Fixes ambiguity with null variable values and default values (fixes #239).

### DIFF
--- a/src/Execution/ValuesResolver.php
+++ b/src/Execution/ValuesResolver.php
@@ -300,15 +300,12 @@ class ValuesResolver
         return new CoercedValue($coercedValues, $errors);
     }
 
-
     /**
      * @param TypeInterface|null $type
      * @return bool
      */
-    protected
-    function isInputType(
-        ?TypeInterface $type
-    ) {
+    protected function isInputType(?TypeInterface $type)
+    {
         return ($type instanceof ScalarType) ||
             ($type instanceof EnumType) ||
             ($type instanceof InputObjectType) ||
@@ -327,13 +324,8 @@ class ValuesResolver
      * @throws GraphQLException
      * @throws InvariantException
      */
-    private
-    function coerceValue(
-        $value,
-        $type,
-        $blameNode,
-        ?Path $path = null
-    ): CoercedValue {
+    private function coerceValue($value, $type, $blameNode, ?Path $path = null): CoercedValue
+    {
         if ($type instanceof NonNullType) {
             return $this->coerceValueForNonNullType($value, $type, $blameNode, $path);
         }
@@ -370,8 +362,7 @@ class ValuesResolver
      * @throws GraphQLException
      * @throws InvariantException
      */
-    protected
-    function coerceValueForNonNullType(
+    protected function coerceValueForNonNullType(
         $value,
         NonNullType $type,
         NodeInterface $blameNode,
@@ -400,8 +391,7 @@ class ValuesResolver
      * @param Path|null     $path
      * @return CoercedValue
      */
-    protected
-    function coerceValueForScalarType(
+    protected function coerceValueForScalarType(
         $value,
         ScalarType $type,
         NodeInterface $blameNode,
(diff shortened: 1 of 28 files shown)


Jalle19:
Can you fix the tests?

crisu83:
Note: I don't think that the null value resolution can be completely fixed so that it works exactly like in the reference implementation because it uses `undefined` for the value when it's not defined and `null` when a value is implicitly set to `null`. It might be possible to implement a fix for this, but I don't see how. @Jalle19 maybe you could take a look and give your 50 cents? Thanks!

crisu83:
@Jalle19 the tests are passing now.

crisu83:
I thought about the `null`/`undefined` problem overnight and I think that we could fix it by introducing a `ConversionResult` class that allows us to represent a value corresponding `undefined` in the JS implementation. Thoughts @Jalle19?

crisu83:
I found and fixed the null conversion issue, the issue was in the test case and not in the library. Duh! 😐

src/Execution/ValuesResolver.php
```diff
@@ -26,13 +27,22 @@
...
+use function array_key_exists;
+use function array_keys;
+use function array_map;
+use function array_merge;
+use function Digia\GraphQL\Test\jsonEncode;
 use function Digia\GraphQL\Util\find;
+use function Digia\GraphQL\Util\invariant;
 use function Digia\GraphQL\Util\keyMap;
 use function Digia\GraphQL\Util\suggestionList;
+use function implode;
```
Jalle19: I would just use \ for these core methods

crisu83: My bad. I let PhpStorm resolve ”issues” on it’s own. 😅

crisu83:
@Jalle19 What is what?

src/Execution/ValuesResolver.php
Jalle19: There's an optin to ignore the root namespace

src/Execution/ValuesResolver.php
```diff
@@ -558,36 +626,38 @@ protected function printPath(?Path $path)
...
-                    'Argument "%s" of required type "%s" was provided the variable "$%s" which was not provided a runtime value.',
-                    $argumentName,
-                    $argumentType,
-                    $variableName
-                ),
-                [$variableNode]
-            );
-        }
-    }
+//    protected static function coerceValueForVariableNode(
```
Jalle19: Can you remove this instead of commenting it out?

src/Execution/ValuesResolver.php
```diff
@@ -26,13 +27,22 @@
 use Digia\GraphQL\Util\ConversionException;
 use Digia\GraphQL\Util\TypeASTConverter;
 use Digia\GraphQL\Util\ValueASTConverter;
+use Exception;
+use Traversable;
+use function array_key_exists;
```
Jalle19: These still shouldn't be here

tests/Functional/Schema/DefinitionPrinterTest.php
```diff
@@ -33,128 +33,127 @@ public function testPrintStringFields(string $source): void
...
-            ],
-            // String field with Int argument with default value
-            [
-                dedent('
-                type Query {
-                  singleField(argOne: Int = 2): String
-                }
-                '),
-            ],
+//            // String field
```
Jalle19: Why are these commented out?

src/Execution/ValuesResolver.php
crisu83: I fixed this and updated the project settings.

src/Execution/ValuesResolver.php
crisu83: Sure. I apparently forgot to remove the unused method.

src/Execution/ValuesResolver.php
crisu83: This has been fixed.

tests/Functional/Schema/DefinitionPrinterTest.php
crisu83: I fixed all the remaining test cases for the definition printer.

PR #351, opened 2019-10-29
Add GraphQL contracts support

SerafimArts:
This is an example of implementing of https://github.com/php-graphql/type-system-contracts into an existing solution which was discussed in https://github.com/webonyx/graphql-php/issues/560

I ask you to make a codereview and draw conclusions based on what we are correcting in the `php-graphql/type-system-contracts` package and `digiaonline/graphql-php`.

Problems I encountered while implementing interfaces:

## Major

1) GraphQL Argument, Field and InputField require the obligatory indication of type. Nulls are not allowed. But the digiaonline implementation allows null as a type.
1.1) GraphQL Contract: https://github.com/php-graphql/type-system-contracts/blob/master/src/Common/TypeAwareInterface.php#L24
1.2) GraphQL Spec: 
1.2.1) https://github.com/graphql/graphql-js/blob/master/src/type/definition.d.ts#L520
1.2.2) https://github.com/graphql/graphql-js/blob/master/src/type/definition.d.ts#L507
1.2.3) https://github.com/graphql/graphql-js/blob/master/src/type/definition.d.ts#L804
1.3) Digia Implementation: https://github.com/digiaonline/graphql-php/blob/master/src/Type/Definition/TypeTrait.php#L15
1.4) Solution: Implementation additional `getNullableType` method: https://github.com/digiaonline/graphql-php/commit/074958e403b579ce24ee56d9767b1c85501110a8

2) Digia implementation does not support repeatable directives. Now the implementation throws an exception if this method is called: https://github.com/SerafimArts/graphql-php/blob/graphql-contracts/src/Type/Definition/Directive.php#L71-L74

## Behavior change

1) Any wrapping types in GraphQL are output and input. This required a change in the logic of the `isInputType` and `isOutputType` functions.
1.1) Solution: https://github.com/digiaonline/graphql-php/commit/d006fabac4917faefc96bd7913d546eaa98fda89

## Minor

1) Incompatibility `SchemaInterface` with `Schema` implementation:
1.1) [SchemaInterface::getPossibleTypes](https://github.com/php-graphql/type-system-contracts/blob/master/src/SchemaInterface.php#L89) and [Schema::getPossibleTypes](https://github.com/digiaonline/graphql-php/blob/master/src/Schema/Schema.php#L236)
1.2) [SchemaInterface::isPossibleType](https://github.com/php-graphql/type-system-contracts/blob/master/src/SchemaInterface.php#L96) and [Schema::isPossibleType](https://github.com/digiaonline/graphql-php/blob/master/src/Schema/Schema.php#L201)
1.3) [SchemaInterface::getQueryType](https://github.com/php-graphql/type-system-contracts/blob/master/src/SchemaInterface.php#L60) and [Schema::getQueryType](https://github.com/digiaonline/graphql-php/blob/master/src/Schema/Schema.php#L139)
1.4) [SchemaInterface::getMutationType](https://github.com/php-graphql/type-system-contracts/blob/master/src/SchemaInterface.php#L65) and [Schema::getMutationType](https://github.com/digiaonline/graphql-php/blob/master/src/Schema/Schema.php#L147)
1.5) [SchemaInterface::getSubscriptionType](https://github.com/php-graphql/type-system-contracts/blob/master/src/SchemaInterface.php#L70) and [Schema::getSubscriptionType](https://github.com/digiaonline/graphql-php/blob/master/src/Schema/Schema.php#L155)

2) Some classes required a change of signatures from `getSomeElements(): array` to `getSomeElements(): iterable` (for example `Schema::getTypeMap(): array` -> `Schema::getTypeMap(): iterable`).

3) Some types requires the addition of `hasSome` methods (for example `Directive::hasLocation(xxx)`).

4) In the current implementation of contracts, an `AbstractTypeInterface` is not an instance of `NamedTypeInterface`. Perhaps it is worth fixing in a contracts package?


src/Schema/Schema.php
```diff
@@ -136,23 +141,23 @@ public function __construct(
     /**
      * @return ObjectType|null
```
Jalle19: Phpdoc should either be updated or removed

src/Schema/Schema.php
```diff
@@ -179,7 +184,7 @@ public function getDirectives(): array
     /**
      * @return array
      */
-    public function getTypeMap(): array
+    public function getTypeMap(): iterable
```
Jalle19: Any particular reason why you chose iterable over array in the contracts?

src/Type/Definition/AbstractTypeInterface.php
```diff
@@ -2,7 +2,10 @@
 
 namespace Digia\GraphQL\Type\Definition;
 
-interface AbstractTypeInterface extends NamedTypeInterface
+use GraphQL\Contracts\TypeSystem\Type\NamedTypeInterface;
+use GraphQL\Contracts\TypeSystem\Type\AbstractTypeInterface as AbstractTypeContract;
```
Jalle19: Why change the name?

src/Type/Definition/ArgumentsTrait.php
```diff
@@ -73,4 +78,22 @@ protected function buildArguments(string $typeName, array $rawArguments): array
...
+        return $this->arguments[$name] ?? null;
+    }
+
+    /**
+     * @param string $name
+     * @return bool
+     */
+    public function hasArgument(string $name): bool
+    {
+        return isset($this->arguments[$name]);
```
Jalle19: Maybe array_key_exists would be better?

src/Schema/Schema.php
SerafimArts: `interable` indicated in the interface and I was afraid that in 7.1 this will not work, because redefinition is only available from 7.4. 

However, later, when I manually checked it, I was convinced that even in php 7.1 this code works: http://sandbox.onlinephpfunctions.com/code/e19dffa7e679c2a9fbaff1592e047eae4c07b19b So I returned everything back. Perhaps in this place I just forgot =)

src/Type/Definition/AbstractTypeInterface.php
SerafimArts: Do you mean `AbstractTypeContract`?

src/Type/Definition/ArgumentsTrait.php
SerafimArts: Yas and No. `isset` is the language opcode and x2-x3 faster than `array_key_exists`. So if guarantees are required that `null` is also a valid argument value, then it should be replaced by:
```
return isset($this->arguments[$name]) || \array_key_exists($name, $this->arguments);
```

SerafimArts:
@Jalle19 I think that now it does not make sense to especially read the code and correct it in details (for example, dock blocks and so on). I would like comments in general on whether such changes will cause problems. I Since its all the same, when there will be a final decision, I will once again look at the code and fix such minor errors.

Of the problems that I am currently observing are those described above, in the description of the PR. In addition, there is a BC. For example, it might make sense to return the existing digiaonline interfaces that I deleted and mark them as `@deprecated`. Then the changes will not be so large.

src/Type/Definition/AbstractTypeInterface.php
Jalle19: Yes

src/Type/Definition/ArgumentsTrait.php
Jalle19: Yeah I was mostly thinking about the "null is valid" case

src/Type/Definition/AbstractTypeInterface.php
SerafimArts: There is `interface AbstractTypeInterface extends AbstractTypeContract {}`

src/Type/Definition/ArgumentsTrait.php
SerafimArts: fixed https://github.com/digiaonline/graphql-php/pull/351/commits/eb13a44b3c9ce08176b6dd061f7b9c7a9193cf6a

Jalle19:
@SerafimArts IMO a BC break is fine, we could merge this and maybe change some other BC breaking things and call it 2.0.0.

@SerafimArts since you seem interested in this library, would you like to have write access? I'm pretty strapped for time myself these days and one of the other original authors left the company some time ago, so resources are kinda scarce on our end at the moment.

src/Schema/Schema.php
Jalle19: Well, if you want to keep `iterable` in the contracts then I guess we'll have to use `iterable` here too.

SerafimArts:
> since you seem interested in this library, would you like to have write access? 

@Jalle19 Well, I have my own (unstable) project: https://github.com/railt/railt In this case, I am only interested in supporting the common parts, to simplify integration because don't have a GraphQL executor.

Adding a type system interfaces will allow me to integrate my parser (https://github.com/railt/railt/tree/master/packages/Parser) and language implementation into various third-party solutions (including your executor) =)

SerafimArts:
---
That is, if support for these interfaces appears in digia, then it will be possible to use my SDL parser without any problems, which contains a lot of all sorts [of cool things](https://github.com/railt/railt/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3ARL%2FSDL+).

Jalle19:
@SerafimArts before I merge, is this good to go in your opinion?

